PR #38, opened 2024-06-10
Fix string plurals

aristath:
## Context
Fixes string plurals in some cases and makes strings easier to translate.

## Quality assurance


classes/class-base.php
aristath, line 1: The `Base::weeks()` method is no longer used after the changes in this PR, so I removed it

jdevalk (approved):

PR #46, opened 2024-07-04
Skip rendering badges that are complete

aristath:
## Context
If all badges are complete, we should not display the separate badge-progress widgets.

* Added a `should_render()` method for widgets. If all badges are complete, then this method will return false.
* Added caching in the `get_badge_details()` methods using a static var to avoid re-calculating things.

## Quality assurance

Fixes #

### DIFF
--- a/classes/widgets/class-badge-content.php
+++ b/classes/widgets/class-badge-content.php
@@ -23,7 +23,7 @@ final class Badge_Content extends Widget {
 
 	/**
 	 * Whether we should render the widget or not.
-	 * 
+	 *
 	 * @return bool
 	 */
 	protected function should_render() {
--- a/classes/widgets/class-badge-content.php
+++ b/classes/widgets/class-badge-content.php
@@ -22,6 +22,22 @@ final class Badge_Content extends Widget {
 	protected $id = 'badge-content';
 
 	/**
+	 * Whether we should render the widget or not.
+	 * 
+	 * @return bool
+	 */
+	protected function should_render() {
+		$details = $this->get_badge_details();
+		if (
+			'awesome-author' === $details['badge']['id']
+			&& 100 === (int) $details['progress']['progress']
+		) {
+			return false;
+		}
+		return true;
+	}
+
+	/**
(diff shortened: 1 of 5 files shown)


classes/widgets/class-badge-content.php
```diff
@@ -21,6 +21,22 @@ final class Badge_Content extends Widget {
...
 
+	/**
+	 * Whether we should render the widget or not.
+	 *
+	 * @return bool
+	 */
+	protected function should_render() {
+		$details = $this->get_badge_details();
+		if (
+			'awesome-author' === $details['badge']['id']
```
jdevalk, line 32: Does it matter which badge it is? If the progress is 100, shouldn't we never render it here?

classes/widgets/class-badge-streak.php
```diff
@@ -22,6 +22,22 @@ final class Badge_Streak extends Widget {
...
 
+	/**
+	 * Whether we should render the widget or not.
+	 *
+	 * @return bool
+	 */
+	protected function should_render() {
+		$details = $this->get_badge_details();
+		if (
+			'super-site-specialist' === $details['badge']['id']
```
jdevalk, line 33: Same question as above.

jdevalk (changes requested):
One question :)

classes/widgets/class-badge-content.php
aristath: If it's 100, then the badge is not pending so internally we move on to the next one.
The only scenario where we could have displayed `100` is if it was the final badge. This is handled in the `get_badge_details()` method:
https://github.com/Emilia-Capital/progress-planner/blob/fdd04d3d7bf064fe7e3b8a8f6e311839622d4b47/classes/widgets/class-badge-content.php#L103-L109

aristath: Hmmm I get what you mean now.
The only scenario where `100` would show, is if it was the final badge.

Makes sense, I'll go ahead and remove the check 👍

jdevalk (approved):

PR #51, opened 2024-07-08
Add a settings popover to exclude post-types from tracking

aristath:
## Context

This PR adds a popover for settings, to allow excluding post-types from tracking.

The settings popover was added to the header, right next to the months selector:

Screenshot 2024-07-08 at 11 11 12 AM

Initially, I wanted to add a setting to select which post-types will be _included_, however that would lead to lots of complications if post-types get added later. With that in mind, it seemed a better option to add a setting to _exclude_ post-types.
To avoid designing a new page etc, it seemed better to reuse the popover API we already implement - especially since we only have 1 option. If in the future we add more options, we may want to rethink that approach

Screenshot 2024-07-08 at 11 42 09 AM

## Summary

* Add settings popover to exclude post-types from tracking.

* Used an AJAX call to save the options, to avoid reloading the page. 

## UI changes

* [x] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

## Documentation

## Quality assurance

### DIFF
--- a/assets/js/settings.js
+++ b/assets/js/settings.js
@@ -1,7 +1,7 @@
 document.getElementById( 'prpl-settings-form' ).addEventListener( 'submit', function( event ) {
 	event.preventDefault();
 	const form = new FormData( this );
-	const data = form.getAll( 'prpl-settings-post-types-exclude[]' );
+	const data = form.getAll( 'prpl-settings-post-types-include[]' );
 
 	// Save the options.
 	progressPlannerAjaxRequest( {
@@ -10,9 +10,9 @@ document.getElementById( 'prpl-settings-form' ).addEventListener( 'submit', func
 			action: 'progress_planner_save_cpt_settings',
 			_ajax_nonce: progressPlanner.nonce,
 			include_post_types: data,
-		}
+		},
 	} );
 
-	document.getElementById( 'submit-exclude-post-types' ).disabled = true;
-	document.getElementById( 'submit-exclude-post-types' ).innerHTML = progressPlanner.l10n.saved;
+	document.getElementById( 'submit-include-post-types' ).disabled = true;
+	document.getElementById( 'submit-include-post-types' ).innerHTML = progressPlanner.l10n.saved;
 } );
--- a/assets/js/settings.js
+++ b/assets/js/settings.js
@@ -1,0 +1,18 @@
+document.getElementById( 'prpl-settings-form' ).addEventListener( 'submit', function( event ) {
+	event.preventDefault();
+	const form = new FormData( this );
+	const data = form.getAll( 'prpl-settings-post-types-exclude[]' );
+
+	// Save the options.
+	progressPlannerAjaxRequest( {
+		url: progressPlanner.ajaxUrl,
+		data: {
+			action: 'progress_planner_save_cpt_settings',
+			_ajax_nonce: progressPlanner.nonce,
+			exclude_post_types: data,
+		}
+	} );
+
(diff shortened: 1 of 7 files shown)


jdevalk:
I've been discussing this with Marieke. We'd like to change it; instead of making it "Exclude",  I think we should make it "Include" and we should default to Posts and Pages "on" and the else "off" by default.

classes/popups/class-settings.php
```diff
@@ -0,0 +1,72 @@
...
+		<h2><?php \esc_html_e( 'Settings', 'progress-planner' ); ?></h2>
+
+		<form id="prpl-settings-form">
+			<h3><?php \esc_html_e( 'Post Types', 'progress-planner' ); ?></h3>
+			<p><?php \esc_html_e( 'Select the post types you want to include in activity scores. This setting will affect which post-type activities get tracked.', 'progress-planner' ); ?></p>
+			<?php foreach ( $post_types as $post_type ) : ?>
+				<label>
+					<input
+						type="checkbox"
+						name="prpl-settings-post-types-exclude[]"
```
jdevalk, line 60: Ok I'm going to be annoying, we should change this to include to prevent future confusion :)

classes/popups/class-settings.php
```diff
@@ -0,0 +1,72 @@
...
+						value="<?php echo \esc_attr( $post_type ); ?>"
+						<?php checked( \in_array( $post_type, $saved_settings, true ) ); ?>
+					/>
+					<?php echo \esc_html( \get_post_type_object( $post_type )->labels->name ); ?>
+				</label>
+			<?php endforeach; ?>
+
+			<button id="submit-exclude-post-types" class="button button-primary"><?php \esc_html_e( 'Save', 'progress-planner' ); ?></button>
```
jdevalk, line 68: We should change this to include to prevent future confusion :)

assets/js/settings.js
```diff
@@ -0,0 +1,18 @@
+document.getElementById( 'prpl-settings-form' ).addEventListener( 'submit', function( event ) {
+	event.preventDefault();
+	const form = new FormData( this );
+	const data = form.getAll( 'prpl-settings-post-types-exclude[]' );
```
jdevalk, line 4: We should change this to include to prevent future confusion :)

assets/js/settings.js
```diff
@@ -0,0 +1,18 @@
...
+	progressPlannerAjaxRequest( {
+		url: progressPlanner.ajaxUrl,
+		data: {
+			action: 'progress_planner_save_cpt_settings',
+			_ajax_nonce: progressPlanner.nonce,
+			include_post_types: data,
+		}
+	} );
+
+	document.getElementById( 'submit-exclude-post-types' ).disabled = true;
```
jdevalk, line 16: We should change this to include to prevent future confusion :)

assets/js/settings.js
```diff
@@ -0,0 +1,18 @@
...
+	document.getElementById( 'submit-exclude-post-types' ).innerHTML = progressPlanner.l10n.saved;
```
jdevalk, line 17: We should change this to include to prevent future confusion :)

jdevalk (changes requested):

jdevalk:
Also; hitting "Save" should trigger a page refresh, as it changes the data that should be displayed.

assets/js/settings.js
aristath: You're right, I missed them in my previous commit. Fixed! 👍

jdevalk (approved):

PR #55, opened 2024-07-08
Fix graph filters

aristath:
## Context

The recently introduced settings for post-types were not affecting the graphs.

This PR adds a filter to allow filtering graphs based on the selected post-types.

IMPORTANT NOTE:
I thought of doing the same for badges, but I think we should leave those as-is... Badges is something that users have already earned and we should not have them un-earn a badge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-= 0.9.5 =
+= - in development =
 
 Fixed:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+= 0.9.5 =
+
+Fixed:
+
+* Post-type filters intruduced in v0.9.4 now also affect the graph results.
+
+= 0.9.4 =
+
+Enhancements:
+
+* Added a setting to include post types, we default to `post` and `page` and you can add others as you wish.
+
+Fixed:
+
+* Completing the last badge wouldn't ever work, fixed.
+* Fixed some bugs around detecting badges being "had".
+* Replaced links to the site with shortlinks, so we can change them as needed without doing a release.
+
 = 0.9.3 =
 
 Security:
--- a/classes/widgets/class-published-content-density.php
+++ b/classes/widgets/class-published-content-density.php
@@ -81,7 +81,25 @@ final class Published_Content_Density extends Widget {
 			],
 			'count_callback' => [ $this, 'count_density' ],
 			'compound'       => false,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+= 0.9.5 =
```
jdevalk, line 1: Let's call this "<version> - in development" until we decide to release,  so we don't make the mistake of adding more releases than one :)

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+
+Fixed:
+
+* Post-type filters intruduced in v0.9.4 now also affect the graph results.
+
+= 0.9.4 =
+
+Enhancements:
+
+* Added a setting to include post types, we default to `post` and `page` and you can add others as you wish.
```
jdevalk, line 11: I see I forgot to set this from the readme, we should copy the text from the readme to here.

classes/widgets/class-published-content-density.php
```diff
@@ -129,14 +146,15 @@ public function get_all_activities_density() {
 		// Get the all-time average.
 		static $density;
 		if ( null === $density ) {
-			$density = $this->count_density(
+			$activities = $this->filter_activities(
 				\progress_planner()->get_query()->query_activities(
```
jdevalk, line 150: Why the `\` in front here? Isn't this in the same namespace?

readme.txt
```diff
@@ -78,6 +78,12 @@ https://youtu.be/e1bmxZYyXFY
 
 == Upgrade Notice ==
 
+= 0.9.5 =
```
jdevalk, line 81: Same remarks as for Changelog.md above.

jdevalk (changes requested):

classes/widgets/class-published-content-density.php
aristath: The `progress_planner()` function is in the main plugin file, so doesn't have a namespace.
It's just shorthand instead of typing `Base::get_instance()`, so we could even remove this function completely 🤔

CHANGELOG.md
aristath: What do you mean? I already copied it...

classes/widgets/class-published-content-density.php
jdevalk: i'm fine with this too.

jdevalk (approved):

PR #58, opened 2024-07-09
Add NPM diff workflow

jdevalk:
## Context

* Allow seeing the changes in package-lock.json more easily.

### DIFF
--- a/.github/workflows/npm-diff.yml
+++ b/.github/workflows/npm-diff.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: NPM lockfile changes
-        uses: codepunkt/npm-lockfile-changes@main
+        uses: emilia-capital/npm-lockfile-changes@main
         with:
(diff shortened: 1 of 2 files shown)


aristath (approved):

.github/workflows/npm-diff.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    # Permission overwrite is required for Dependabot PRs, see "Common issues" below.
+    permissions:
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: NPM lockfile changes
+        uses: emilia-capital/npm-lockfile-changes@main
```
aristath, line 15: Getting an error that this repo doesn't exist.
Is it a repo we should create, or change this line to something else?

aristath (changes requested):

jdevalk:
Ah forget it, I remember now that I need to find another way to do this.

PR #61, opened 2024-09-03
🚧  WIP - Add suggested tasks

aristath:
## Context

*

## Summary

*

*

#### Relevant test scenarios
* [ ] Changes should be tested with the browser console open
* [ ] Changes should be tested on different posts/pages/taxonomies/custom post types/custom taxonomies
* [ ] Changes should be tested on different editors (Block/Classic/Elementor/other)
* [ ] Changes should be tested on different browsers
* [ ] Changes should be tested on multisite

## UI changes

* [ ] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

## Documentation

## Quality assurance

Fixes #


jdevalk:
Nice!

Played a bit, some minor things, not even considering design yet: 

- when adding to the to-do list, that shouldn't "cross it", I think, only when it's actually completed.
- when it's "completed", I can't expand the tasks any more, they no longer have any explanation?

aristath:
> when adding to the to-do list, that shouldn't "cross it", I think, only when it's actually completed.

When adding an item to the to-do list, I assume we won't want that item to be present in both places, so I just `dismiss` it from the list. 

> when it's "completed", I can't expand the tasks any more, they no longer have any explanation?

The initial implementation for `dismiss` is currently just "crossing" the title and hiding the details via CSS. I'm sure we'll find a better implementation for the `dismiss` buttons. 
Perhaps "dismiss" should actually remove the item from the list, but for now this simple method was making debugging easier for me, and it was easier to visualize what happens 😄 

We should definitely figure out the details of how this feature will function and what the user experience should be 👍

aristath:
Just a note: The reason this PR is getting HUGE because suggested tasks required refactoring other things, including a refactor of the dashboard layout to correspond with the new mockup

PR #63, opened 2024-09-06
Tweaks & improvements

aristath:
## Context
I was doing a code review today of the plugin, so this PR is just fixing & improving some things I came across.

## Summary

* Fix i18n issues for strings contained in JS files
* Minor code cleanup & readability tweaks

## Quality assurance

Fixes #

### DIFF
--- a/assets/js/todo.js
+++ b/assets/js/todo.js
@@ -23,7 +23,10 @@ jQuery( document ).ready( function () {
 				.attr( 'aria-label', content );
 			jQuery( this )
 				.find( '.trash' )
-				.attr( 'aria-label', `Delete task '${ content }'` );
+				.attr(
+					'aria-label',
+					progressPlannerTodo.i18n.taskDelete.replace( '%s', content )
+				);
 		} );
 
 		if ( todoList.length === 0 ) {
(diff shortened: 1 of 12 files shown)


assets/js/todo.js
aristath, line 1: Fixed i18n issues in this file

classes/activities/class-content.php
```diff
@@ -88,13 +88,16 @@ public function get_points_on_publish_date() {
...
-			$points *= Base::$points_config['content']['word-multipliers'][350];
-		} elseif ( $words > 100 ) {
-			$points *= Base::$points_config['content']['word-multipliers'][100];
+			return (int) ( $points * $multipliers[1000] );
+		}
+		if ( $words > 350 ) {
+			return (int) ( $points * $multipliers[350] );
+		}
+		if ( $words > 100 ) {
+			return (int) ( $points * $multipliers[100] );
```
jdevalk, line 100: Should this be a `switch`? Think that's easier to read.

aristath: We can't use `>` comparisons with a `switch`...

jdevalk: Hmm true, although we could do:

```php
switch (true) {
    case $words > 1000:
        do stuff;
        break;

    case $words > 350:
        do stuff;
        break;

    case $words > 100:
        do stuff;
        break;

    default:
        do stuff;
        break;
}
```

But doesn't add that much either.

jdevalk (approved):

PR #67, opened 2024-09-10
Add an onboarding tour

jdevalk:
This adds an onboarding tour, which is prompted to start after the user has finished the onboarding and scanning the content has completed.

Prompt to start the tour

It also adds an `[i]` button next to the settings button that always starts the onboarding tour:

Button to start the tour

The tour itself goes through different sections of the plugin:

Plugin tour showing a plugin section

## Technical decisions

- Added a hook `do_action( 'progress_planner_admin_page_header_before' );` to add content above the header. Used this with the Playground Show / Hide onboarding feature too.
- Made a generic CSS class for `prpl-button-primary` to have nice buttons and not repeat styles.
- Created a `prplStartTour` function in `tour.js`, this might need review as you might see a better way to do this.


assets/js/scan-posts.js
```diff
@@ -31,10 +31,10 @@ const progressPlannerTriggerScan = () => {
...
 				'progress-planner-scan-progress'
 			).style.display = 'none';
-			window.location.href = window.location.href.replace(
-				'&content-scan',
-				''
-			);
+
+			window.location.href =
+				window.location.href + '&content-scan-finished=true';
+
```
aristath, line 37: Should this be like this perhaps?
```suggestion

			window.location.href =
				window.location.href.replace(
				'&content-scan',
				''
			) + '&content-scan-finished=true';

```
The current code will not remove the `content-scan` param, should we?

assets/css/admin.css
```diff
@@ -703,9 +715,36 @@ button.prpl-info-icon {
...
+	border-radius: var(--prpl-border-radius);
+}
+
+.prpl-wrap input.prpl-button-primary:hover,
+.prpl-wrap button.prpl-button-primary:hover,
+.prpl-wrap input.prpl-button-primary:focus,
+.prpl-wrap button.prpl-button-primary:focus {
+	text-decoration: underline;
+	box-shadow: 3px 3px 10px var(--prpl-color-accent-red);
+}
```
aristath, line 742:
```suggestion

	&:hover,
	&:focus {
		text-decoration: underline;
		box-shadow: 3px 3px 10px var(--prpl-color-accent-red);
	}
}
```

views/admin-page-header.php
```diff
@@ -15,7 +15,20 @@
...
+<?php if ( isset( $_GET['content-scan-finished'] ) ) : // phpcs:ignore WordPress.Security ?>
+	<div class="prpl-widget-wrapper prpl-top-notice">
+	<div class="inner-content">
+		<h1><?php esc_html_e( "We've successfully scanned your site's content.", 'progress-planner' ); ?></h1>
+		<p><?php esc_html_e( 'Would you like a tour of the plugin now?', 'progress-planner' ); ?></p>
+			<div id="prpl-onboarding-submit-grid-wrapper">
+				<button class="prpl-button-primary" onclick="prplStartTour()"><?php esc_html_e( 'Yes, let\'s go!', 'progress-planner' ); ?></button>
+			</div>
+		</div>
+	</div>
```
aristath, line 29: Should this notice also have a `Skip tour` button?

aristath: Also, when the tour finishes, we should hide the notice

assets/js/scan-posts.js
jdevalk: I've never seen the `content-scan` in the URL...

views/admin-page-header.php
jdevalk: Did the latter, will add a close button.

assets/js/scan-posts.js
aristath: Yeah, it's just an easter-egg, allowing you to reset and re-scan your content 😉

views/admin-page-header.php
jdevalk: Added close buttons, to the Playground notice too.

assets/js/tour.js
aristath, line 1: Just to be safe, I think it would be best to prefix all our constants and functions and avoid conflicts in case another plugin does something similar:
* `driver`
* `driverObj`
* `makePopoverBackdropTransparent`

aristath (changes requested):
Looks good! Just a nitpick tweak in `tour.js` to avoid conflicts 👍

assets/js/tour.js
jdevalk: Good catch. Actually will prefix the CSS class used too.

jdevalk: Done.

aristath (approved):

PR #69, opened 2024-09-10
Switch Playground to use wp.ajax too

jdevalk:
Does what it says on the tin _and_ fixes the onboarding switch in the playground.

Please test in Playground whether everything works as expected.


aristath (approved):
Looks good, works as expected 👍

PR #72, opened 2024-09-17
Make welcome box a fullscreen popover

aristath:
## Context

* Change the welcome box to be a full-screen popover.


classes/class-onboard.php
```diff
@@ -72,77 +70,79 @@ public static function the_form() {
...
 				)
 				?>
 			</div>
-			<label>
-				<span class="prpl-label-content">
-					<?php \esc_html_e( 'First Name', 'progress-planner' ); ?>
-				</span>
+			<br>
+			<div>
+				<label style="display:block;">
```
jdevalk, line 80: Should we add this style to the stylesheet?

jdevalk: (and the one below)

assets/js/onboard.js
```diff
@@ -75,13 +75,71 @@ const progressPlannerOnboardCall = ( data ) => {
...
+						} );
+					document.querySelector(
+						'#prpl-onboarding-form .prpl-form-fields'
+					).style.display = 'none';
+					document.querySelector(
+						'#prpl-onboarding-form .prpl-button-primary'
+					).style.display = 'none';
+					document.querySelector(
+						'#prpl-onboarding-form .prpl-button-secondary--no-email'
+					).style.display = 'block';
```
jdevalk, line 97: This doesn't look very DRY, especially as we do the same below. Can we dry it up a bit?

jdevalk (changes requested):
Some minor comments

jdevalk (approved):

PR #77, opened 2024-10-10
Change how blog feed transient is saved

ilicfilip:
The idea is to make make our Dashboard page accessible to the user even if new blog feed can't be fetched from our site for some reason (ie timeout occurs).

In case that happens user will see "stale" blog posts and next try to fetch "fresh" posts will be delayed by 5 minutes.

The change also handles migration from currently saved transient to the new format, that includes expiration time so our API shouldn't be hammered once users update the plugin.

### DIFF
--- a/classes/widgets/class-whats-new.php
+++ b/classes/widgets/class-whats-new.php
@@ -66,31 +66,58 @@ final class Whats_New extends Widget {
 	 * @return array
 	 */
 	public function get_blog_feed() {
-		$feed = \get_site_transient( 'progress_planner_blog_feed_with_images' );
-		if ( false === $feed ) {
+		$feed_data = \get_site_transient( 'progress_planner_blog_feed_with_images' );
+
+		// Migrate old feed to new format.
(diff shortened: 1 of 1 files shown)


classes/widgets/class-whats-new.php
```diff
@@ -66,31 +66,58 @@ public function the_content() {
 	 * @return array
 	 */
 	public function get_blog_feed() {
-		$feed = \get_site_transient( 'progress_planner_blog_feed_with_images' );
-		if ( false === $feed ) {
+		$feed_data = \get_site_transient( 'progress_planner_blog_feed_with_images' );
```
aristath, line 69: Could we move the `progress_planner_blog_feed_with_images` in a `const` in the object? 🤔 
This way we can avoid repeating it all over the place, and we can change `_site_transient_timeout_progress_planner_blog_feed_with_images` to `'_site_transient_timeout_' . self::TRANSIENT_NAME` to make things clearer

ilicfilip: Sure, done.

aristath (approved):

PR #82, opened 2024-10-11
Add a db-version option and implement methods to upgrade the database

### LINKED ISSUE #80: Consider adding a DB table version
In https://github.com/Emilia-Capital/progress-planner/issues/78 we changed DB column type by making a check which makes additional DB query on every load.

To avoid adding similar queries in the future we should consider adding a DB table version. Then we can do table "migration", for example changing column types or data, only if version is lower than that "migration" requires.

If we do it, we should make sure that multiple "migrations" run in correct order (in case user updates from a much older version).

aristath:
## Context
Future-proofing:
* Added a `progress_planner_db_version` option
* Added a `Query::maybe_upgrade` method to run updates
* Added an upgrade routine for the `data_id` column

* Routines for upgrade methods are named `upgrade_{DATE}`, where `DATE` is a numeric value year-month-day (example: `20241011` if the upgrade routine was written in October 11, 2024). This way we can ensure all upgrades run sequentially without getting too involved in finding version numbers for the db-upgrades etc.

Fixes #80

### DIFF
--- a/classes/class-query.php
+++ b/classes/class-query.php
@@ -48,6 +48,7 @@ class Query {
 	 */
 	private function __construct() {
 		$this->create_tables();
+		$this->maybe_upgrade();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


ilicfilip (approved):
I have just changed one var name (to be more consistent with the foreach loop above it), the rest looks good - tried to break it while testing but without success.

PR #108, opened 2024-11-06
Improve cache when fetching lessons

ilicfilip:
## Context
Due to the way how currently our code works fetching lessons from API is done on every `init` (I get it even on front end, not just in Dashboard).

If request is successful the response is cached, but if it is not our server will be hammered. To prevent that I cached empty array with 5 minutes lifetime in case that request fails.

Another problem is that `$url` can be filtered, but the cache key is never changed - meaning that even if $url is changed user will keep getting lessons which were fetched from old `$url` until the cache expires (which can be up to a week).

I used `$url` in order to make `$cache_key` unique and left it unchanged in order to make possible (future) debugging easier.

## Quality assurance

### DIFF
--- a/classes/class-lessons.php
+++ b/classes/class-lessons.php
@@ -53,7 +53,7 @@ class Lessons {
 			'https://progressplanner.com/wp-json/progress-planner-saas/v1/free-lessons'
 		);
 
-		$cache_key = 'lessons-' . $url;
+		$cache_key = 'lessons-' . md5( $url );
 
 		$cached = \progress_planner()->get_cache()->get( $cache_key );
 		if ( is_array( $cached ) ) {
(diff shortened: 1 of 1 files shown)


classes/class-lessons.php
```diff
@@ -42,37 +42,45 @@ public function get_items() {
...
+		 * Filter the endpoint url for the lessons.
+		 *
+		 * @param string $endpoint The endpoint url.
+		 */
+		$url = apply_filters(
+			'progress_planner_lessons_endpoint',
+			'https://progressplanner.com/wp-json/progress-planner-saas/v1/free-lessons'
+		);
+
+		$cache_key = 'lessons-' . $url;
```
aristath, line 56: It would probably be safer if we `md5()` the URL...

```suggestion
		$cache_key = 'lessons-' . md5( $url );
```

ilicfilip: Sure, I left it just because of the debugging - any hash will do the job.

aristath (approved):

PR #126, opened 2024-11-19
Add a new `prpl-suggested-task` web component

aristath:
## Context

* Convert suggested tasks to use a web-component for items.

## Test instructions

* Check that the items get properly rendered, both in the admin page as well as the main dashboard widget
* Check that the info/snooze buttons and actions still work properly


ilicfilip:
I have moved listeners to the web component as well, to me it feels like they belong there - except that, it looks good to me (tested widget both on PP and Dashboard page).

PR #130, opened 2024-11-19
Local suggested tasks

ilicfilip:
Refactoring local suggested tasks


classes/class-suggested-tasks.php
```diff
@@ -144,6 +194,12 @@ public function mark_task_as_celebrated( $task_id ) {
 	 * @return void
 	 */
 	public function maybe_celebrate_tasks() {
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- We're not processing any data.
+		if ( ! is_admin() || ! isset( $_GET['page'] ) || $_GET['page'] !== 'progress-planner' ) {
+			return; // WIP: Only celebrate tasks in the admin area.
+		}
+
```
aristath, line 202: We can probably use `\ get_current_screen()` here...

```suggestion
		$current_screen = \get_current_screen();
		if ( ! $current_screen || 'toplevel_page_progress-planner' !== $current_screen->id ) {
			return; // WIP: Only celebrate tasks in the admin area.
		}

```

ilicfilip: 👍 

Yes, that was just a WIP - that is unfished WIP code which was actually ran on every page (including front end)

Reason why I was seeing confetti only occasionally (since `pending_celebration` option was cleared on other pages as well).

aristath (approved):
LGTM 👍

PR #160, opened 2024-12-06
Refactor suggested tasks

ilicfilip:


aristath (commented):
Classes in the `Progress_Planner\Suggested_Tasks\Local_Tasks\Providers` namespace all contain `Provider` in their name. I think we can probably remove that to simplify their names... (example: `Update_Content_Provider` -> `Update_Content`). The `Provider` is implied from the namespace

ilicfilip:
Yes, I moved them to separate directory in last commit, so it makes sense now.

Also, we can probably name them differently - for example:
```
Create_Content
Update_Content
Update_Core
```

to:

```
Content_Create
Content_Update
Core_Update
```

so they are grouped when we add more tasks which are related.

Didnt want to do it before checking with you as that will also mean that we need to rename getters in our codebase.

aristath (approved):

PR #163, opened 2024-12-06
Fix duplicated update core tasks

ilicfilip:
## Context
We had a bug reported today by @tacoverdo , after Core update confetti were triggered on every page load.

We had a bug which caused that update core task with the same ID can be added multiple times, resulting the task being at the same time marked as `pending_celebration` and `completed`.

Until now we looped through the suggested tasks `<ul>` (in the Widget), looking for the tasks which are mark as pending for celebration and marking them as completed. The problem was that also in JS we filter out completed tasks from rendering in the list, which resulted in confetti being triggered over and over.

I have fixed this by preventing update-core task with the same ID to be added twice. Furthermore I moved transitioning tasks from `pending_celebration` to `completed` on page load (in PHP), so it is no more relevant if the task is actually rendered in the list or not.

Besides this I found a bug with results returned from `\wp_get_update_data`, in [method used for generating tasks](https://github.com/Emilia-Capital/progress-planner/blob/0a4c78610d68605ae2daec43ae220bf04ef4c809/classes/suggested-tasks/local-tasks/class-update-core.php#L43) it was working fine but in the [method for evaluating tasks](https://github.com/Emilia-Capital/progress-planner/blob/0a4c78610d68605ae2daec43ae220bf04ef4c809/classes/suggested-tasks/local-tasks/class-update-core.php#L22) it was not as it was called too early. The [get_core_updates](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/update.php#L921) function wasnt defined, which resulted in always returning 0 for available Core updates (marking the task as completed although nothing was updated).

## Summary

Fix duplicated update core tasks

## Quality assurance


aristath (approved):

aristath:
@ilicfilip this looks good, there's just a merge conflict that needs to be resolved. After you resolve the conflict, feel free to merge the PR 👍

PR #171, opened 2024-12-09
Add a popover for plugin-deactivation feedback

aristath:


jdevalk:
Can we make the backdrop black? Gives a bit more focus to the popover.

jdevalk (approved):

PR #173, opened 2024-12-10
Design tweaks & monthly badges init from the activation date

ilicfilip:
## Context
This PR syncs PP page with latest design tweaks (mostly spacing, icons and hover states) and also changes the way monthly badges are inited.

It was needed so we init badges starting from the plugin activation date.

## Quality assurance


views/page-widgets/suggested-tasks.php
```diff
@@ -30,21 +52,24 @@
 </div>
 
 <hr>
```
aristath, line 54: Nitpick: Indentation

aristath (approved):
Looks good and works as expected. I only had a nitpick regarding indentation, but otherwise looks good 👍

PR #181, opened 2024-12-11
Auto-detect page-types in the settings page

aristath:
This PR will attempt to auto-detect any existing posts for the homepage, contact, fax & about-us pages, and pre-populate the option in our settings page.


ilicfilip (approved):

ilicfilip:
I have only changed the DB query to fetch only the ID (as that is the only thing we need).

PR #184, opened 2024-12-12
Define remote url constant as a (global) constant

ilicfilip:
## Context

We have our remote server URL defined in many places, it is defined for every use case separately.

This PR defines it as a global PHP constant which makes it consistent. 
Also `is_defined` allows us to easily change URL in case we want to set up server instance locally or in a staging environment.

## Quality assurance

### DIFF
--- a/progress-planner.php
+++ b/progress-planner.php
@@ -26,6 +26,10 @@ define( 'PROGRESS_PLANNER_FILE', __FILE__ );
 define( 'PROGRESS_PLANNER_DIR', __DIR__ );
 define( 'PROGRESS_PLANNER_URL', untrailingslashit( plugin_dir_url( __FILE__ ) ) );
 
+if ( ! defined( 'PROGRESS_PLANNER_REMOTE_SERVER_ROOT_URL' ) ) {
+	define( 'PROGRESS_PLANNER_REMOTE_SERVER_ROOT_URL', 'https://progressplanner.com' );
+}
+
 /**
  * Autoload classes.
  */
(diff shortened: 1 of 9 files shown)


progress-planner.php
```diff
@@ -26,6 +26,10 @@
 define( 'PROGRESS_PLANNER_DIR', __DIR__ );
 define( 'PROGRESS_PLANNER_URL', untrailingslashit( plugin_dir_url( __FILE__ ) ) );
 
+if ( ! defined( 'PROGRESS_PLANNER_REMOTE_SERVER_ROOT_URL' ) ) {
+	define( 'PROGRESS_PLANNER_REMOTE_SERVER_ROOT_URL', 'https://progressplanner.com' );
+}
```
aristath, line 31: Instead of doing this, could we do it in the `Base` class, so we can call `\Progress_Planner\Base::REMOTE_SERVER_ROOT_URL`? I would prefer it as opposed to a global const...

aristath (changes requested):

progress-planner.php
aristath: Or even a method so we can call `\progress_planner()->get_remote_server_root_url()`, and we can add a const check in there...
so something like
```php
public function get_remote_server_root_url() {
  return defined( 'PROGRESS_PLANNER_REMOTE_SERVER_ROOT_URL' )
    ? PROGRESS_PLANNER_REMOTE_SERVER_ROOT_URL
    : 'https://progressplanner.com';
}

ilicfilip: We can, the only important is that is defined in one place and that we can override it

ilicfilip: @aristath , updated

`classes/widgets/class-latest-badge.php` needed to be changed, since we can't concatenate strings in a property declaration.

aristath (approved):

PR #186, opened 2024-12-13
Save settings page task

ilicfilip:
## Context
Implements https://github.com/Emilia-Capital/progress-planner/issues/154

## Quality assurance

### DIFF
--- a/views/admin-page-settings.php
+++ b/views/admin-page-settings.php
@@ -50,6 +50,61 @@ if ( ! defined( 'ABSPATH' ) ) {
 			</div>
 		</div>
 
+		<div class="prpl-column">
+			<div class="prpl-widget-wrapper">
+				<h2 class="prpl-settings-section-license">
+					<span>
+						<?php \esc_html_e( 'Licenses', 'progress-planner' ); ?>
+					</span>
+				</h2>
+				<div class="prpl-license-keys-wrapper">
+					<?php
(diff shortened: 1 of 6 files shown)


views/page-widgets/suggested-tasks.php
```diff
@@ -28,7 +28,10 @@ class="prpl-info-icon"
 			</button>
 
 			<div class="prpl-tooltip">
-				<?php \esc_html_e( 'Collect 7 points by completing the tasks and earn this month\'s badge.', 'progress-planner' ); ?>
+				<?php
+				/* translators: %d: The target points. */
+				printf( \esc_html__( 'Collect %d points by completing the tasks and earn this month\'s badge.', 'progress-planner' ), (int) \Progress_Planner\Badges\Monthly::TARGET_POINTS );
+				?>
```
aristath, line 34: Good catch! 👍

views/admin-page-settings.php
```diff
@@ -50,6 +50,61 @@
 			</div>
 		</div>
 
+		<div class="prpl-column">
```
aristath, line 53: Until we implement licensing, perhaps this should be hidden? 

```suggestion
		<div class="prpl-column" style="display:none;">
```

aristath (changes requested):

views/admin-page-settings.php
ilicfilip: Done

aristath (approved):

PR #187, opened 2024-12-13
Move checklist before writer's block section

jdevalk:
Switches the lesson sections around, or should, at least.


aristath (approved):
Tested and confirmed, works as expected 👍

PR #194, opened 2024-12-19
Remove unnecessary filters

aristath:
There won't be a separate PRO plugin, so we can remove a lot of the filters.

### DIFF
--- a/classes/suggested-tasks/class-remote-tasks.php
+++ b/classes/suggested-tasks/class-remote-tasks.php
@@ -106,6 +106,6 @@ class Remote_Tasks {
 				$url
 			);
 		}
-		return \apply_filters( 'progress_planner_suggested_tasks_remote_api_endpoint', $url );
+		return $url;
 	}
 }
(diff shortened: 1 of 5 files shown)


classes/suggested-tasks/class-remote-tasks.php
```diff
@@ -106,6 +106,6 @@ protected function get_api_endpoint() {
 				$url
 			);
 		}
-		return \apply_filters( 'progress_planner_suggested_tasks_remote_api_endpoint', $url );
```
ilicfilip, line 109: I think we added this filter also to allow 3rd party integration (ie hosts), but we can always add it back when that is thought through.

ilicfilip (approved):

classes/suggested-tasks/class-remote-tasks.php
aristath: Yeah, if I'm not mistaken we'll be implementing an alternative method using our site as a proxy. Still under discussion and not urgent at this stage, but we don't need the filter for now.

PR #196, opened 2024-12-19
Check user capability for suggested tasks

ilicfilip:
## Context

The issue is that tasks were always added to the Suggested tasks list, even if current user doesnt have capability to perform them. Good examples are `Update Core` and `Fill settings page` tasks. 

So for start, tasks links are added conditionally.
Second, I decided to skip adding tasks in such cases to the list until we decide do we want it or not - perhaps we would like to change task description in that case, ie `Ask site admin perform all updates`

Also one weird edge case was that admin could actually peform the task, but if editor was logged in at the same time it could happen that confetti was triggered for him (although editor didnt do anything). That is why the condition to skip "pending_celebration" tasks was added in the `classes/widgets/class-suggested-tasks.php` was added.

I have made a several tests, but needs to be tested a bit more before merging since we're close to the release.

## Quality assurance

* [] I have tested this code to the best of my abilities.

### DIFF
--- a/classes/suggested-tasks/local-tasks/providers/class-content-abstract.php
+++ b/classes/suggested-tasks/local-tasks/providers/class-content-abstract.php
@@ -10,7 +10,14 @@ namespace Progress_Planner\Suggested_Tasks\Local_Tasks\Providers;
 /**
  * Add tasks for content updates.
  */
-abstract class Content_Abstract {
+abstract class Content_Abstract extends Local_Tasks_Abstract {
+
+	/**
+	 * The capability required to perform the task.
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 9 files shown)


aristath (changes requested):
Instead of repeating the `capability_required` method in all classes, perhaps we should put it in a parent class and then simply define a class prop?

So something like this:
```php
protected function capability_required() {
	return $this->capability
		? \current_user_can( $this->capability )
		: true;
}
```

ilicfilip:
Agree, just we need new parent class just for that one method - so I left it until we have more to abstract.

But I will add it now.

aristath (approved):

PR #200, opened 2024-12-23
Clean up local tasks

ilicfilip:
## Context

For some type of suggested tasks (create new short / long post, update core) we use year and week as part of their ID - this way we can generate new tasks of the same type periodically and also prevent user to do the same task infinite number of times during the same week (ie to create long / short repeatedly).

For https://github.com/Emilia-Capital/progress-planner/pull/163 I changed format of the `update-core `task to include year and week, but I haven't updated all the checks properly - so the `update-core` task is added to the local (suggested) tasks list again when the next week starts. It is not visible to the user (it is not displayed in the widget) but upon completion 2 update tasks would be struck through and user would get 2 points.

On the other side, create new short / long post had a check if completed task(s) was generated within the current week, so points are awarded correctly, but with the same problem of task(s) of the same type being internally added when new week starts.

I have looked into several ways on how to fix this and the least invasive and IMO the best approach is to:
- add a check for completed `update-core` tasks if they were generated in the current week (like we had for create post tasks)
- add a cleanup method which will remove uncompleted tasks which were generated in the previous week (this runs once a day)

## Summary

Remove duplicate weekly suggested tasks.

## Quality assurance

* [x] I have added unit tests to verify the code works as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+= 1.0.2 =
+
+Fixed:
+
+* Duplicate weekly suggested tasks.
+
 = 1.0.1 =
 
 Fixed:
--- a/classes/suggested-tasks/class-local-tasks-manager.php
+++ b/classes/suggested-tasks/class-local-tasks-manager.php
@@ -240,6 +240,13 @@ class Local_Tasks_Manager {
 	 * @return void
 	 */
 	public function cleanup_pending_tasks() {
+
+		$cleanup_recently_performed = \progress_planner()->get_cache()->get( 'cleanup_pending_tasks' );
+
+		if ( $cleanup_recently_performed ) {
+			return;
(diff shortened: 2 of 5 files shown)


ilicfilip:
One more thing, probably this is a good place to discuss - For every task we have unique ID, for example:

- For create posts or update core it is the year and week suffix
- For update posts it is the post ID
- For remote tasks it is the ID of the task on the server side

This makes sense since ID should be unique, but we also have some tasks which IMO dont require it, like "Fill the settings page' (which has year and week suffix as well).

Currently it does not make the difference, but in the future we will have more similar tasks (ie "Yoast: Set organization logo") and if the task ID contains such suffix it means user can (theoretically) unset & set a setting to earn points every week.

jdevalk (approved):

PR #203, opened 2024-12-25
Fallback for remote host images

ilicfilip:
## Context

Since 1.0 version we pull most of the images from the remote server. In case our server is not accessible for some reason or an user is using our plugin locally without internet access (for example while traveling) PP page will look a bit broken:

<details>
<summary>PP page</summary>
Screenshot 2024-12-25 at 13 15 53
</details>

<details>
<summary>Badges popover</summary>
Screenshot 2024-12-25 at 13 15 59
</details>

This PR adds a fallback in case images failed to load:
<details>
<summary>PP page with fallback images</summary>
Screenshot 2024-12-25 at 13 47 56
</details>

Placeholders needs to be replaced, but those are correct dimensions - there are 2 sizes, smaller one for the badges and the bigger one is used as a default size (currently that is only "Latest new badge").

Blog post images are added differently, as a background (using CSS ) to an element which has width and height set, so I added one of the shades of grey we have defined as a background color.

Definitely not a priority, but nice to have.

## Summary

Added fallback for remote host images.

## Quality assurance


classes/class-base.php
```diff
@@ -144,6 +144,18 @@ public function get_remote_server_root_url() {
...
 	}
 
+	/**
+	 * Get the placeholder image URL.
+	 *
+	 * @return array<string, string>
+	 */
+	public function get_placeholder_image_url() {
+		return [
+			'default' => 'https://placehold.co/1200x675',
```
jdevalk, line 154: Hmm this seems like a good idea, but not really from a GDPR perspective 😅  can we replicate this easily in WP or otherwise?

jdevalk (changes requested):

classes/class-base.php
jdevalk: Let's just do this?

```php
function get_placeholder_svg( $width, $height ) {
	return 'data:image/svg+xml;base64,' . base64_encode( sprintf( '<svg width="%1$d" height="%2$d" xmlns="http://www.w3.org/2000/svg"><rect x="2" y="2" width="%3$d" height="%4$d" style="fill:#F6F5FB;stroke:#534786;stroke-width:2"/><text x="50%%" y="50%%" font-size="20" text-anchor="middle" alignment-baseline="middle" font-family="monospace" fill="#534786">%1$d&#215;%2$d</text></svg>', $width, $height, ( $width - 4 ), ( $height - 4 ) ) );
}
```

jdevalk (approved):

PR #205, opened 2024-12-27
Fix get_default_page_id_by_type

ilicfilip:
## Context
Recently we added a feature which should pre-select page for the page type in the "Your Pages" section on the Settings page. It works by searching through existing pages' titles, any found matches for specific page type are compared to matches found for other page types and any overlapping results are excluded. What is left is returned as a suggestion.

There is a bug there, which @tacoverdo described in https://github.com/Emilia-Capital/progress-planner/issues/201 and I believe it was introduced when then feature was refactored. 

I think that [$posts](https://github.com/Emilia-Capital/progress-planner/blob/1a7e4dd3019c277bc52aa710013e23ec680b1a9b/classes/class-page-types.php#L316) here shouldn't be always set to `contact pages` and `return empty( $posts ) ? 0 : $posts[0];` (on line 332) will execute after the first loop iteration (which means other page types wont be checked).

I refactored `get_default_page_id_by_type` to be simpler and to me it looks like it does the same thing, but I will need another pair of eyes since I didn't implement this and I might be wrong.

Also I changed [homepage subarray](https://github.com/Emilia-Capital/progress-planner/blob/1a7e4dd3019c277bc52aa710013e23ec680b1a9b/classes/class-page-types.php#L299) to be just home page ID, instead of the WP_Post object, to be consistent with the others. As far I saw it doesnt have to be an object and it might be set to it because in initial implementation Post objects were used for other subarrays as well (but then we changed it so `get_posts_by_title` method returns just IDs to be lighter).

## Summary

Fixes suggested pages for "Your pages" section on Settings page.

## Quality assurance

Fixes https://github.com/Emilia-Capital/progress-planner/issues/201


jdevalk:
Should I see a change in behavior here on Playground @ilicfilip ?

Maybe @tacoverdo you can look at this one as well?

ilicfilip:
The thing here is that bug depends on the existing pages, since we search through them to detect best suggestions, so I dont think this can be replicated on Playground.

I can double check with @tacoverdo, for example apply the fix on his site directly, but the best would be that @aristath takes a look at it as well since he initially implemented this feature.

jdevalk:
Ok I'll wait on @aristath to review this one!

tacoverdo:
I can replicate the behavior as follows on Playground: 

- Start playground via the link above
- Create a page called 'contact'. 
- Go to the PP settings.
- Try to change the setting for 'contact page' from "I have this page" to "I don't have this page yet"
- Click Save 

For me, it still jumps back to "I have this page".

ilicfilip:
Checking, but that might be kind of what we want - to suggest a page to a user if selection wasn't made previously.

The actual bug which was fixed, and it was more obvious in your [initial issue](https://github.com/Emilia-Capital/progress-planner/issues/201), is that suggestions were wrong. In your video you were seeing the issue on the FAQ page, but the page which was suggested was "Contact" page (which shouldn't be suggested in that context at all).

ilicfilip:
@tacoverdo , what I said above is correct - that is the way we offer a suggestion to a user when we hasn't selected a page previously or marked it as "My site doesn't need this page".

What would we like to do here, to suggest a page (and make selections) only when user loads the settings page for the first time? If pages are pre-selected, is it clear that user needs to be save the Settings page in order for selection to take an effect?

aristath (approved):

PR #217, opened 2025-01-17
Fix missing SVN

jdevalk:
The new Ubuntu image that runs actions doesn't contain `svn`, so we need to install that before using it.

This fixes the integration tests, and brings the deploy and asset deploy tasks to `develop` where this has also been fixed.


jdevalk:
@aristath please review - I think this fix works and we need to fix the integration test failures separately?

aristath (approved):

PR #218, opened 2025-01-20
Remove email-subscription requirement to get pro

aristath:


assets/js/settings.js
ilicfilip, line 1: This code is unrelated to email-subscription requirement, but I just double checked and it is not needed any more.

ilicfilip (approved):

jdevalk (approved):

jdevalk:
@aristath @ilicfilip why are the tests failing? Not sure I understand.

PR #220, opened 2025-01-22
More badges tests

ilicfilip:
## Context
Just some more tests, this time checks if badges array is correctly inited.

## Quality assurance


ilicfilip:
@aristath , the last commit here fixes the failing integration tests (locally they worked).

Transient name was changed in our code, but haven't in tests (thus wrong cache transient was set and terms weren't created).

aristath (approved):

PR #234, opened 2025-02-03
Allow filtering which widgets to display in the dashboard

aristath:
Because maybe I'm not interested in having a TODO list, and for me it's just visual noise


ilicfilip:
Tested, no issues.

If we decide to merge this one we sh
Screenshot 2025-02-03 at 13 34 41
ould update the tour as well, steps 4 & 5:

jdevalk:
I think we'll close this one in favor of drag & drop, which we'll add a bit later.

PR #235, opened 2025-02-03
New tasks

jdevalk:
## Context

* This adds 3 new suggested tasks on top of the one in #207. 

## Summary

* Adds a suggested task to disable the display of PHP errors.
* Adds a suggested task to delete the default "Hello world!" post.
* Adds a suggested task to delete the default "Sample page" page.

* I moved `get_provider_type()` to the abstract.
* I think we should abstract this code quite a bit more, there's quite a bit of duplication here now.

Closes #207.


tacoverdo:
When testing this pull, I noticed that there is no celebration when "completing" the settings page task. However, it is removed from the recommendations. 

https://github.com/user-attachments/assets/d2195fb0-4308-4f5f-9ed5-20e8b5543893

tacoverdo:
I will get the recommendation to remove the sample page, even when I've completely changed its contents and I've turned it into a valuable page. 
Could it be that we're just checking for postID 1 to exist, instead of comparing the current content to the sample content?

ilicfilip:
@tacoverdo , this PR is still a bit buggy (I pushing fixes and changes for the most of the day) so it's somewhat expected that something is not working (I will test the issue you described).

As for the sample content, we're testing for the slug - did you change it?

tacoverdo:
@ilicfilip In my previous test, I did not change the slug, just the content. So I tested again now and changed the slug of the sample page. It completed the task "Delete sample page". I don't think that's an action many users will take when getting the instruction to _delete_ a sample page, but I guess it serves the purpose. I mean, how many users would change the slug without changing the page? 

That said, because the recommendation is "Delete sample page", would a user expect that clicking it automagically deletes that page, or that they're taken to the edit page for the sample page? My expectation would be the former, but I can also see why it would be a bad idea to thrash the page unseen.

jdevalk:
Added a comment to close #207 when we merge this one too.

PR #237, opened 2025-02-05
Check for translatable slug and title for the sample page

aristath:
## Context

The `sample-page` slug and `Sample page` titles are translatable in WP-Core. For example, I can see that the slug is `voorbeeld-pagina` in Dutch.

This PR attempts to check for the translated slug. If that fails, it then checks for a page with the `Sample page` title (again, translatable).

### DIFF
--- a/classes/suggested-tasks/local-tasks/providers/class-hello-world.php
+++ b/classes/suggested-tasks/local-tasks/providers/class-hello-world.php
@@ -91,6 +91,29 @@ class Hello_World extends Local_Tasks_Abstract {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


ilicfilip:
`get_tasks_to_inject` and `get_task_details` also needed to be updated, so I did that and since we had repeated code then I moved getting `Sample Page` to separate method.

One thing I changed there `'post_status'  => 'publish'` (instead of `any`) because in original implementation task was marked as completed as soon it was moved to trash.

I also applied the same for `Hello World!` post and tested both with Serbian localization.

PR #238, opened 2025-02-05
Reduce code duplication

aristath:
* move `get_tasks_to_inject` to the parent
* Use `$this->get_provider_type()` instead of `static::TYPE`
* Use `$this->get_provider_id()` instead of `static::ID`

### DIFF
--- a/classes/suggested-tasks/local-tasks/providers/class-content-create.php
+++ b/classes/suggested-tasks/local-tasks/providers/class-content-create.php
@@ -149,7 +149,11 @@ class Content_Create extends Content_Abstract {
 	 *
 	 * @return array
 	 */
-	public function get_task_details( $task_id ) {
+	public function get_task_details( $task_id = '' ) {
+
+		if ( ! $task_id ) {
+			return [];
+		}
 
 		$data = $this->get_data_from_task_id( $task_id );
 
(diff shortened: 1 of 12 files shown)


ilicfilip (commented):
I reviewed the changes, there are couple of bugs which can be solved easily but to fix all the `move get_tasks_to_inject to the parent ` change would need to be reverted (or adjusted) - main issue is that those tasks were being celebrated over and over again.

Let's leave this PR open for now and we can review once the changes involved with the new-tasks branch are done.

ilicfilip:
@aristath , I have pushed simplification for the one time tasks, defining that kind of tasks should be [much simpler](https://github.com/ProgressPlanner/progress-planner/blob/4e8fbb3974cfaa3e6bd9d037e717f8d2442832c2/classes/suggested-tasks/local-tasks/providers/class-core-blogdescription.php).

Tomorrow I will check if we can do something similar for the repetitive tasks.

ilicfilip:
@aristath , can you take a look at this one?

I created 2 new abstract classes, for one time tasks (site icon, blog description, saving settings, debug display, hello world and sample page) and for repetitive tasks (currently only core update but any other similar should be added easily).

Create / update posts are a bit specific and their logic didn't quite fit into what we do with other tasks, so I left them as they were.

jdevalk (approved):

PR #240, opened 2025-02-05
Display 1 suggested task per type, but with JS filtering

ilicfilip:
## Context
This PR improves implementation from the `jdv/new-tasks` branch, where we changed how suggested tasks are displayed in the widget (1 task from every available type).

Now that limit filtering is moved to Javascript, like it was before, which allows us to display new task to the user immediately after task is snoozed.

We still need to decide if we want to display all available types to the user, makes sense to do so but previously we had a limit of 5 tasks on PP page and 3 tasks on WP Dashboard page.

## Quality assurance

### DIFF
--- a/assets/js/widgets/suggested-tasks.js
+++ b/assets/js/widgets/suggested-tasks.js
@@ -1,4 +1,4 @@
-/* global customElements, progressPlannerSuggestedTasks, confetti, prplDocumentReady, progressPlannerSuggestedTask */
+/* global customElements, progressPlannerSuggestedTasks, confetti, prplDocumentReady */
 
 /**
  * Count the number of items in the list.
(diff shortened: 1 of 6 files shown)


assets/js/widgets/suggested-tasks.js
```diff
@@ -191,48 +204,41 @@ const prplStrikeCompletedTasks = () => {
 		document
 			.querySelectorAll( '.prpl-suggested-task-celebrated' )
 			.forEach( ( item ) => {
-				const taskId = item.getAttribute( 'data-task-id' );
-
-				const request = wp.ajax.post(
-					'progress_planner_suggested_task_action',
```
aristath, line 197: Why are we removing this action from here? Won't that break the implementation? 🤔

ilicfilip: We dont need it :)

Since the endless confetti issue, marking tasks as completed is done in PHP [here](https://github.com/ProgressPlanner/progress-planner/blob/5aef23be09ea07f317f62bfd29ddda073647748f/classes/widgets/class-suggested-tasks.php#L134).

That was just left over from the old implementation and even it's PHP callback was doing nothing (other than confirming that request was successful) so I removed it both.

aristath (approved):

PR #241, opened 2025-02-06
Remove admin notices in the plugin's dashboard page

### LINKED ISSUE #221: Remove admin notices from plugin settings page
## Is your feature request related to a problem? Please describe.
Admin notices can severely clutter the admin experience in WordPress. Some plugins are overly aggressive with them, and show their admin notices on plugin admin pages as well. To keep the Progress Planner admin experience clean, we should remove admin notices from our Dashboard and settings pages. 

## Describe the solution you'd like
A potential solution to this is Accessibility Checker's implementation, which can be found at https://github.com/equalizedigital/accessibility-checker/pull/835/files

## Why do you think this feature is something we should consider for this plugin?
Because it will keep the admin experience for Progress Planner cleaner, and thus more useable.

aristath:
Fixes #221

### DIFF
--- a/classes/admin/class-page.php
+++ b/classes/admin/class-page.php
@@ -220,7 +220,17 @@ class Page {
 	 */
 	public function remove_admin_notices() {
 		$current_screen = \get_current_screen();
-		if ( ! $current_screen || 'toplevel_page_progress-planner' !== $current_screen->id ) {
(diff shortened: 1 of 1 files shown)


ilicfilip (approved):
LGTM and tested as well.

PR #242, opened 2025-02-06
Add option to redirect on login

aristath:
## Summary

* Added option to redirect users to the Progress Planner dashboard after login.

## Test instructions

* As an administrator, go to the plugin options and enable the new option
* Logout and login again. You should be redirected to the plugin dashboard.
* Logout, and login as a different user, one that doesn't have the `manage_options` capability. You should NOT be redirected.
* Confirm that disabling the option stops the login redirect

Fixes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Enhancements:
 * Improvements to the REST-API endpoint for getting stats.
 * Added a plugin-deactivation feedback form.
 * Removed celebration for "Perform all updates" if it was done by WordPress' automatic update.
+* Added option to redirect users to the Progress Planner dashboard after login.
 
 = 1.0.3 =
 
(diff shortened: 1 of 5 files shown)


aristath:
> I have a suggestion, and I might be nitpicking, but how about making this setting per user (using user meta for saving it)?

Excellent point!
Done 👍

aristath:
> When saving settings page I get this PHP warning:
> 
> `PHP Warning: Undefined array key "id" in /Users/filip/Valet/planner/htdocs/wp-content/plugins/progress-planner/classes/admin/class-page-settings.php on line 162`

Not related to this PR.
Looking into it, but shouldn't be a blocker for this one

aristath:
Fixed the PHP notice in https://github.com/ProgressPlanner/progress-planner/commit/df1ca81f19c86d6ae087164c90243d9bf42c30c0 👍

ilicfilip (approved):
I cleaned up one more instance of `$this->get_settings()->get( 'redirect_on_login' )`, tested and it works.

PR #246, opened 2025-02-06
Confetti hearts

ilicfilip:
For testing GET parameters can be used, for example:

`wp-admin/admin.php?page=progress-planner&start_date=2025-02-05&end_date=2025-02-07`

### DIFF
--- a/classes/widgets/class-suggested-tasks.php
+++ b/classes/widgets/class-suggested-tasks.php
@@ -145,11 +145,12 @@ final class Suggested_Tasks extends Widget {
 
 		// Check if current date is between Feb 12-16.
 		$confetti_options = [];
-		$current_date     = \gmdate( 'Y-m-d' );
+		$year             = '2025';
+		$current_date     = $year . '-' . \gmdate( 'm-d' );
 
 		// TODO: GET params just for testing.
-		$start_date = isset( $_GET['start_date'] ) ? \sanitize_text_field( \wp_unslash( $_GET['start_date'] ) ) : '2025-02-12';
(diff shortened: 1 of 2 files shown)


jdevalk (approved):

classes/widgets/class-suggested-tasks.php
```diff
@@ -143,6 +143,31 @@ public function enqueue_scripts() {
 			$max_items_per_type['pending_celebration'] = 99;
 		}
 
+		// Check if current date is between Feb 12-16.
+		$confetti_options = [];
+		$current_date     = \gmdate( 'Y-m-d' );
+
+		// TODO: GET params just for testing.
+		$start_date = isset( $_GET['start_date'] ) ? \sanitize_text_field( \wp_unslash( $_GET['start_date'] ) ) : '2025-02-12';
```
jdevalk, line 151: Maybe we shouldn't hardcode the year here?

classes/widgets/class-suggested-tasks.php
```diff
@@ -145,11 +145,12 @@ public function enqueue_scripts() {
 
 		// Check if current date is between Feb 12-16.
 		$confetti_options = [];
-		$current_date     = \gmdate( 'Y-m-d' );
+		$year             = '2025';
```
jdevalk, line 148: Could we just make this `date('Y')` ? means we don't have to do it again next year :)

jdevalk (approved):

classes/widgets/class-suggested-tasks.php
jdevalk: Haha I said *shouldn't* :)

PR #255, opened 2025-02-11
Dont inject already completed local tasks

ilicfilip:
## Context
This issue is a bit tricky and it surfaced today on @goodplansam site, most likely because it has a lot of completed tasks.
The reported issue was that there was a suggested task for updating an old page which can't be dismissed (reappears after the page reload).

It has 2 parts, first one is related to creating new `review-post` tasks, we haven't checked if the task with the same id (basically the same task) was already completed.

I also added another check directly in the `inject_tasks` method to prevent this from happening in the future, since that is the place where all tasks are fetched from all task providers. The downside is that we do the same check in multiple places now, in every task provider and here, so for performance reasons we should probably discuss it and agree what would be the best place for it (including the remote tasks as well).

The second part, and this is why it was hard to catch, was that we also filter tasks in Javascript before adding them to the widget (we filter out completed, snoozed and tasks already in the list). That works fine in most cases, but in this specific case it revealed that we were [splicing the array while we were looping through it.](https://github.com/ProgressPlanner/progress-planner/blob/develop/assets/js/widgets/suggested-tasks.js#L50C2-L62C6) which led to described issue.

## Summary

Improved checks for suggested 'review post' tasks.

## Quality assurance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Under the hood:
 
 * Improved suggested tasks completion conditions.
+* Improved checks for suggested 'review post' tasks.
 
 = 1.0.4 =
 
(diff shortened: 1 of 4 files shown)


aristath (approved):
awesome catch!

PR #256, opened 2025-02-12
Minor coding tweaks & readability improvements

aristath:


classes/class-suggested-tasks.php
```diff
@@ -318,7 +318,8 @@ public function remove_task_from( $status, $task_id ) {
 		$option[ $status ] = isset( $option[ $status ] )
 			? $option[ $status ]
 			: [];
-		$remove_index      = false;
+
+		$remove_index = \array_search( $task_id, $option[ $status ], true );
```
ilicfilip, line 322: With this change we will loop through 'snoozed' array twice, once here and the second time in the IF block (but first loop will always return false).

That is the reason why it was the way the array_search was in the ELSE block.

classes/class-suggested-tasks.php
```diff
@@ -538,27 +533,23 @@ public function check_task_condition( $condition ) {
...
-			[
-				'type'    => 'completed',
-				'task_id' => $task_id,
-			]
-		) ) {
-			return true;
-		}
-
-		return false;
+		return (
```
ilicfilip, line 536: I have put it like this at first, but then I thought 2 IF blocks would be more readable to others 😄

PR #257, opened 2025-02-12
Define script dependencies in scripts files

aristath:
## Context
This PR simplifies the way we define script dependencies.
Instead of adding things in PHP, we can now define script dependencies in the JS files themselves by using the file headers.
Example:
```js
/*
 * Dependencies: foo, bar, baz
 */
```


ilicfilip (approved):
nice one!

PR #259, opened 2025-02-12
Include WP_DEBUG in "is debug display enabled" task

ilicfilip:
## Context
Until now we were checking if `WP_DEBUG_DISPLAY` constant was defined and set to true. But this constant is only taken into account if `WP_DEBUG` is set to true as well, so we're updating the check to cover that case.

Link to docs: https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/#wp_debug_display

Strictly, if both of this constants are set to false displaying errors will fallback to what is PHP system setting.
On production sites usually errors will be hidden, but mentioning it in case this is tested locally because for me errors were showing even with constant set to false (until I changed PHP ini settings to mimic production environment).

## Quality assurance

Fixes #

### DIFF
--- a/classes/suggested-tasks/local-tasks/providers/class-debug-display.php
+++ b/classes/suggested-tasks/local-tasks/providers/class-debug-display.php
@@ -37,7 +37,12 @@ class Debug_Display extends Local_OneTime_Tasks_Abstract {
 		 * For WP_DEBUG_DISPLAY to do anything, WP_DEBUG must be enabled (true).
 		 * link: https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/#wp_debug_display
 		 */
-		return defined( 'WP_DEBUG' ) && ( false === WP_DEBUG || ( WP_DEBUG && ( ! defined( 'WP_DEBUG_DISPLAY' ) || ! WP_DEBUG_DISPLAY ) ) ) ? true : false; // @phpstan-ignore-line
+		return (
+			defined( 'WP_DEBUG' ) && (
+				false === WP_DEBUG ||
+				( ! defined( 'WP_DEBUG_DISPLAY' ) || ! WP_DEBUG_DISPLAY )
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


classes/suggested-tasks/local-tasks/providers/class-debug-display.php
```diff
@@ -27,12 +27,17 @@ class Debug_Display extends Local_OneTime_Tasks_Abstract {
...
 	 *
 	 * @return bool
 	 */
 	public function check_task_condition() {
-		return ( ! defined( 'WP_DEBUG_DISPLAY' ) || ! WP_DEBUG_DISPLAY ) ? true : false;
+		/**
+		 * For WP_DEBUG_DISPLAY to do anything, WP_DEBUG must be enabled (true).
+		 * link: https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/#wp_debug_display
+		 */
+		return defined( 'WP_DEBUG' ) && ( false === WP_DEBUG || ( WP_DEBUG && ( ! defined( 'WP_DEBUG_DISPLAY' ) || ! WP_DEBUG_DISPLAY ) ) ) ? true : false; // @phpstan-ignore-line
```
aristath, line 40: Not a blocker, just a suggestion, minor readability tweak:
* Splits long condition to multiple lines
* Removes one of the nested `WP_DEBUG` checks
* Removes the `? true : false;` part since the condition already returns a bool result
```suggestion
		return (
			defined( 'WP_DEBUG' ) && (
				false === WP_DEBUG ||
				( ! defined( 'WP_DEBUG_DISPLAY' ) || ! WP_DEBUG_DISPLAY )
			)
		);
```

aristath (approved):

classes/suggested-tasks/local-tasks/providers/class-debug-display.php
ilicfilip: Sure!

I usually leave the `? true : false;` part for readability and phpstan complained about doing `Removes one of the nested WP_DEBUG checks` (which I also left for readability since it is literal "translation" of what is said in the WP docs), but we can change it if is clear enough what is done.

PR #260, opened 2025-02-13
Add new task to disable comments.

aristath:


ilicfilip:
Confetti weren't triggering (dependency was missing) when this task was dismissed, I added it and improved check so it will be automatically added for any new dismissible local task type. 

Also I found something which might be a bug:

- Fresh install, one comment and task gets added
- I add 10 more comments, so this task condition `return 10 >= \wp_count_comments()->approved;` is not longer true
- I open PP page and task gets celebrated

aristath:
@ilicfilip thanks for the check!

I changed the condition, so now the task will only get added if `default_comment_status` is not `closed`, and there are fewer than 10 comments.

ilicfilip (approved):

ilicfilip:
@tacoverdo , @irisgue , I am tagging you to let you know that this task probably needs a description update and a page on our site.

PR #261, opened 2025-02-13
Reverse the logic for conditions checks

aristath:
* Replaces the `check_task_condition` methods with `should_add_task`, reversing the logic
* Add new `check_task_condition` methods for backwards-compatibility reasons in case a 3rd-party plugin was registering tasks

### DIFF
--- a/classes/class-suggested-tasks.php
+++ b/classes/class-suggested-tasks.php
@@ -458,7 +458,7 @@ class Suggested_Tasks {
 	 *
 	 * @return bool
 	 */
-	public function maybe_add_task( $condition ) {
+	public function should_add_task( $condition ) {
 		$parsed_condition = \wp_parse_args(
 			$condition,
 			[
@@ -531,7 +531,7 @@ class Suggested_Tasks {
 	 * @return bool
 	 */
 	public function check_task_condition( $condition ) {
-		return ! $this->maybe_add_task( $condition );
+		return ! $this->should_add_task( $condition );
 	}
 
 	/**
@@ -545,7 +545,7 @@ class Suggested_Tasks {
 
 		return (
 			// Check if the task was pending celebration.
-			false === $this->maybe_add_task(
+			false === $this->should_add_task(
 				[
 					'type'    => 'pending_celebration',
 					'task_id' => $task_id,
@@ -553,7 +553,7 @@ class Suggested_Tasks {
 			)
 			||
 			// Check if the task was completed.
-			false === $this->maybe_add_task(
+			false === $this->should_add_task(
 				[
 					'type'    => 'completed',
 					'task_id' => $task_id,
--- a/classes/class-suggested-tasks.php
+++ b/classes/class-suggested-tasks.php
@@ -458,7 +458,7 @@ class Suggested_Tasks {
 	 *
 	 * @return bool
 	 */
-	public function should_add_task( $condition ) {
+	public function check_task_condition( $condition ) {
 		$parsed_condition = \wp_parse_args(
 			$condition,
 			[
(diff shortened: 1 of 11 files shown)


classes/class-suggested-tasks.php
```diff
@@ -458,7 +458,7 @@ private function maybe_unsnooze_tasks() {
 	 *
 	 * @return bool
 	 */
-	public function check_task_condition( $condition ) {
+	public function should_add_task( $condition ) {
```
ilicfilip, line 461: I believe changes in this file should be reverted (at least the renaming).

It is the method with the same but in a different class, which does something else - checks if given task is completed, snoozed and etc (not if task should be added).

We discussed it on Zoom call, it was added months ago and I am fine with renaming it something having more sense now (like check_task_status or similar) but different from `should_add_task` to avoid confusion in the future again.

As far I can see that method is called only in [here](https://github.com/ProgressPlanner/progress-planner/blob/145cb5a55c2ecb80bfdcda527a68463f88a29380/classes/suggested-tasks/local-tasks/providers/class-content-create.php#L45) and [here](https://github.com/ProgressPlanner/progress-planner/blob/145cb5a55c2ecb80bfdcda527a68463f88a29380/classes/suggested-tasks/local-tasks/providers/class-content-create.php#L70) in create content class.

ilicfilip (changes requested):

classes/suggested-tasks/local-tasks/providers/class-hello-world.php
```diff
@@ -46,10 +46,8 @@ class Hello_World extends Local_OneTime_Tasks_Abstract {
 	 *
 	 * @return bool
 	 */
-	public function check_task_condition() {
-		$hello_world = $this->get_sample_post();
-
-		return null === $hello_world ? true : false;
+	public function should_add_task() {
+		return null !== $this->get_sample_post();
```
ilicfilip, line 50: Checking the code I assume that `should_add_task` should return true if task needs to be added.

If that is the case then this one is wrong, we want to add task when sample_post is `null` - but `null !== null `will return `false`

classes/suggested-tasks/local-tasks/providers/class-sample-page.php
```diff
@@ -41,15 +41,12 @@ class Sample_Page extends Local_OneTime_Tasks_Abstract {
...
+	 * Check if the task should be added.
 	 *
 	 * @return bool
 	 */
-	public function check_task_condition() {
-		$sample_page = $this->get_sample_page();
-
-		return null === $sample_page ? true : false;
+	public function should_add_task() {
+		return null !== $this->get_sample_page();
```
ilicfilip, line 49: Same as for sample post ([class-hello-world.php](https://github.com/ProgressPlanner/progress-planner/pull/261/files#diff-fbfe0124846bcbc237bd07b95a9afa7d977d3c2fa19eafff51a70ba4babdf31b)) check, `null !== null` will result in `false`.

classes/suggested-tasks/local-tasks/providers/class-debug-display.php
```diff
@@ -27,22 +27,12 @@ class Debug_Display extends Local_OneTime_Tasks_Abstract {
...
-		 * link: https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/#wp_debug_display
-		 */
-		return (
-			defined( 'WP_DEBUG' ) && (
-				false === WP_DEBUG ||
-				( ! defined( 'WP_DEBUG_DISPLAY' ) || ! WP_DEBUG_DISPLAY )
-			)
-		);
+	public function should_add_task() {
+		return defined( 'WP_DEBUG' ) && defined( 'WP_DEBUG_DISPLAY' ) && WP_DEBUG_DISPLAY;
```
ilicfilip, line 35: I believe this one is wrong, since:

define( 'WP_DEBUG', false);
define('WP_DEBUG_DISPLAY', true);

will result in true (meaning that task should be added?).

But in order for debug display to be enabled both `WP_DEBUG` and `WP_DEBUG_DISPLAY` need to be set to `true` (this is why we updated the condition in https://github.com/ProgressPlanner/progress-planner/pull/259 )

ilicfilip (changes requested):

classes/suggested-tasks/local-tasks/providers/class-hello-world.php
ilicfilip: My mistake, was going through checks and reversing logic - this check is correct.

classes/suggested-tasks/local-tasks/providers/class-debug-display.php
aristath: Good catch, fixed 👍

ilicfilip (approved):

PR #265, opened 2025-02-13
New task: Update PHP version

ilicfilip:
## Context
New task for updating PHP version

## Quality assurance

Fixes https://github.com/ProgressPlanner/progress-planner-pro/issues/82

### DIFF
--- a/classes/suggested-tasks/local-tasks/providers/class-php-version.php
+++ b/classes/suggested-tasks/local-tasks/providers/class-php-version.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Add tasks for settings saved.
+ *
+ * @package Progress_Planner
+ */
+
+namespace Progress_Planner\Suggested_Tasks\Local_Tasks\Providers;
+
+/**
+ * Add tasks for settings saved.
+ */
+class Php_Version extends Local_OneTime_Tasks_Abstract {
+
+	/**
+	 * The provider type.
+	 *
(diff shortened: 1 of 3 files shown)


classes/suggested-tasks/local-tasks/providers/class-php-version.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @package Progress_Planner
+ */
+
+namespace Progress_Planner\Suggested_Tasks\Local_Tasks\Providers;
+
+/**
+ * Add tasks for settings saved.
+ */
+class Php_Version extends Local_OneTime_Tasks_Abstract {
```
aristath, line 13: Should we rename this to `PHP_Version`? I _think_ PHP should be capitalized since it's an acronym...

aristath (approved):

classes/suggested-tasks/local-tasks/providers/class-php-version.php
ilicfilip: I did that at first, but then changed to not get phpcs error and be sure that it works with autoloader / our magic __get methods.

aristath: makes sense 👍

ilicfilip:
@tacoverdo , @irisgue , I am tagging you to let you know that this task probably needs a description update and a page on our site.

PR #269, opened 2025-02-14
Focus on options when there is a recommendation

aristath:
## Summary

This PR adds visual indicators to the settings that users should change in order to gain points based on Ravi's recommendations:

* When we visit an admin page that has settings related to Ravi's recommendations, we show an indicator next to these settings, showing how many points they get
* On the header of the page, we show how many points they've gotten from that page, and the total they can get
* If they go to a page directly by clicking on one of Ravi's recommendations, then we focus on that option changing the indicator to make it stand out some more, and we scroll to that setting's position on the page
* If they complete that task, we change the indicator to a check mark and increase the points count on the page header
 
https://github.com/user-attachments/assets/e85fd0fe-9b40-4617-99a2-ef35853ed6d9

## Quality assurance

Fixes #


jdevalk:
OK I love this, I have one comment though:

CleanShot 2025-02-18 at 13 40 10

aristath:
The only viable option I could come up with is placing the indicator in the beginning, are we OK with something like this?

Screenshot 2025-02-19 at 8 06 44 AM

jdevalk:
Hmm the badge changes the positioning of the words behind it, not ideal.

Also; it’s now unclear what I should do here. I think we need to provide a bit more context OR put that thing in front of the right radio option…

aristath:
Sure, we can put it in front of the right radio option... But which one _is_  the right one?

jdevalk:
I’d say month and name @jonoalderson you agree? Or do we need more explanation here?

aristath:
Converted the PR to draft while I refactor a few things here to make it easier and allow for more customization

aristath:
Refactored the implementation so now the pills get added via JS instead of CSS. This allowed us to add multiple pills, and also fixed some positioning issues. The pills are now inline instead of forced to have `position:absolute`, which I feel is a lot better.
This also means that for the permalinks option I was able to add the pills right next to the options that we think they should select 👍

jdevalk:
@aristath I've added a `scroll-margin-top` to `.prpl-element-awards-points-icon-wrapper` to give a bit of breathing room to the element when we scroll to it. Let me know if you think there's any issue with that?

assets/js/focus-element.js
```diff
@@ -0,0 +1,91 @@
+/* global progressPlannerFocusElement */
+
+const prplGetIndicatorElement = ( content, taskId ) => {
+	// Create an <img> element.
+	const imgEl = document.createElement( 'img' );
+	imgEl.src =
+		progressPlannerFocusElement.base_url +
+		'/assets/images/icon_progress_planner.svg';
+	imgEl.alt = '';
```
jdevalk, line 9: Can't we come up with a proper `alt` attribute for this?

aristath:
> @aristath I've added a `scroll-margin-top` to `.prpl-element-awards-points-icon-wrapper` to give a bit of breathing room to the element when we scroll to it. Let me know if you think there's any issue with that?

I love it!
No issues whatsoever 👍

jdevalk (changes requested):

classes/suggested-tasks/local-tasks/providers/one-time/class-disable-comments.php
```diff
@@ -79,6 +79,10 @@ public function get_task_details( $task_id = '' ) {
 				),
 				(int) \wp_count_comments()->approved
 			) . '</p>',
+			'link_setting' => [
+				'hook'   => 'options-discussion.php',
+				'iconEl' => 'tr th:has(+td label[for="default_comment_status"])',
```
ilicfilip, line 84: Shouldn't this line be changed to: `'iconEl' => 'label[for="default_comment_status"]',` ?

Now looks like indicator is positioned wrongly (at least for me)?
Screenshot 2025-02-20 at 10 29 35


Also I didn't get a point although I turned off the comments.
(my guess is probably since I am testing on fresh WP instance, which only has 1 comment and thus this task is not yet added)

ilicfilip (approved):

assets/js/focus-element.js
jdevalk: It's not decorative though, is it?

classes/suggested-tasks/local-tasks/providers/one-time/class-disable-comments.php
aristath: All fixed

PR #270, opened 2025-02-14
Add Debug Tools to the plugin

ilicfilip:
## Context
We are merging our Debug Tools into the plugin.

Debug tools can be enabled by:

- define( 'PRPL_DEBUG', true );
- or adding an `prpl_debug` entry to the `wp_options` table

## Quality assurance


aristath:
Could we also update the option in https://github.com/ProgressPlanner/progress-planner/blob/develop/classes/class-playground.php#L31 so it's enabled in the playground automatically?

PR #271, opened 2025-02-14
AAA Option Optimizer

ilicfilip:
## Context
This PR adds a suggested task when high number of autoloaded options is detected and it instructs user to install AAA Option Optimizer plugin.

## Quality assurance


classes/suggested-tasks/local-tasks/providers/one-time/class-reduce-autoloaded-options.php
```diff
@@ -0,0 +1,146 @@
...
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+
+		$this->title       = \esc_html__( 'Reduce number of autoloaded options', 'progress-planner' );
+		$this->description = sprintf(
+			// translators: %d is the number of autoloaded options.
+			\esc_html__( 'There are %d autoloaded options. If you don\'t need them, consider disabling them by installing the "AAA Option Optimizer" plugin.', 'progress-planner' ),
```
aristath, line 75: I doubt normal users will understand or know whether they _need_ autoloading options or not, and we should make it clear that this is a performance improvement
```suggestion
			\esc_html__( 'There are %d autoloaded options. Consider installing the "AAA Option Optimizer" plugin to improve performance.', 'progress-planner' ),
```

ilicfilip: I agree, wording was just a placeholder during WIP and it needs to be improved - that is the one for @tacoverdo and @irisgue , if we decide to merge it 😄

ilicfilip:
It was a suggestion, at one point it looked like we're close to merge it but now we went into different direction.

We can both leave it and close it if creates clutter (not hard to re-implement).

jdevalk (approved):

PR #276, opened 2025-02-18
Fix ajax request callbacks

ilicfilip:
## Context
We do AJAX requests from one function, to which we pass data, URL, success and fail callbacks.
In that function we're registering new callback on the `http.onreadystatechange` event, success callback was executing fine but fail callback was triggered on every state change - mainly on when http.readyState values were 2 & 3 (headers received & response downloading), which is wrong.

I have changed it so fail callback is triggered only when request is done, but response code is not 200.

I noticed this while working on new Onboarding flow, scanning posts was triggered twice and there were multiple `'Failed to scan posts. Retrying...'` messages in the console, which looked suspicious.

## Summary

## Quality assurance

### DIFF
--- a/assets/js/ajax-request.js
+++ b/assets/js/ajax-request.js
@@ -34,14 +34,19 @@ const progressPlannerAjaxRequest = ( {
 				return http.response;
 			}
 		}
-		if ( http.readyState === 4 && http.status === 200 ) {
-			return successAction
-				? successAction( response )
+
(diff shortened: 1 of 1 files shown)


jdevalk (approved):

jdevalk:
I'll let @aristath review too, but this looks good to me.

aristath (approved):

PR #285, opened 2025-02-21
Refactor constants and move task id generation to a separate (inheritable) function

jdevalk:
- move all class properties to constants
- make those constants `protected`
- simplify task ID retrieval

### DIFF
--- a/classes/suggested-tasks/local-tasks/providers/repetitive/class-core-update.php
+++ b/classes/suggested-tasks/local-tasks/providers/repetitive/class-core-update.php
@@ -58,7 +58,7 @@ class Core_Update extends Repetitive {
 	public function get_task_details( $task_id ) {
 
 		if ( ! $task_id ) {
-			$task_id = $this->get_provider_id() . '-' . \gmdate( 'YW' );
+			$task_id = $this->get_task_id();
 		}
 
 		return [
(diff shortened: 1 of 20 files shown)


ilicfilip (approved):

classes/suggested-tasks/local-tasks/providers/repetitive/class-core-update.php
```diff
@@ -13,34 +13,26 @@
  */
 class Core_Update extends Repetitive {
 
-	/**
-	 * The capability required to perform the task.
-	 *
-	 * @var string
-	 */
-	protected $capability = 'update_core';
```
aristath, line 21: Why remove this one? 🤔

aristath (approved):
Tested, works as expected and code looks good. 
There's one thing I did not understand though, added a comment

classes/suggested-tasks/local-tasks/providers/repetitive/class-core-update.php
ilicfilip: It's on line 35

aristath: Ah, you're right, I missed that 👍

PR #286, opened 2025-02-21
Interactive task type

ilicfilip:
## Context
Implementation of https://github.com/ProgressPlanner/progress-planner-pro/issues/86

This PR adds example of interactive task, when clicking on a task user gets a popover which let him tests if sending emails is working on his site. If user confirms that sending email is working task will be celebrated, if not user gets an instruction on how to fix it.

Technically, idea is to add interactive part as a custom web component so we can change it any way we want without touching the rest of the code base.

The PHP class for this task type should also contain all relevant code, in this case it is a AJAX callback which sends an email.

## Summary
Added task for checking if sending emails is working.

## Quality assurance


assets/js/tasks/sending-email.js
aristath, line 1: Should be `assets/js/web-components/prpl-email-test-popup.js`

ilicfilip: One of the things I wanted to discuss, to create a separate subfolder for task components, for example: `assets/js/web-components/tasks/prpl-email-test-popup.js` ?

Not to be mixed with the other web components, since we will probably add more interactive tasks

aristath: I assumed that if we add more interactive tasks, we'll probably abstract the component a bit, instead of copy-paste-and-tweak? 🤔

ilicfilip: Sure!

I am currently looking into not autoloading task components once the task is completed.

aristath: What about `assets/js/web-components/prpl-task-email-test-popup.js`?
This way they will all be grouped together since they will start with `prpl-task-*`

ilicfilip: Done <3

ilicfilip:
@Mijke , @tacoverdo here is the screen recording of the current state of the Interactive task (checking if email sending is working):

https://github.com/user-attachments/assets/a5ab07c7-7087-4cb7-8f98-8440ed6acdac

aristath (approved):

PR #290, opened 2025-02-21
Ravi image used for confetti

ilicfilip:
## Context

During the last meeting it was mentioned that it would be nice to use Ravi image as a confetti shape, this PR implements that.

Also it sets the confetti origin to the Suggested Tasks widget, something that @Mijke mentioned that it would be nice to change if possible.

## Quality assurance


jdevalk:
OK I have:

* Added the other badges (`content`, `maintenance` and `month`) that the user has collected or, if none, the one they're currently working on;
* Removed the `width` and `height` from the images, it's not needed and otherwise makes them look weird;
* Made the images a lot bigger, so you can actually recognize them :)

I'd love for @Mijke, @tacoverdo and @irisgue to look at this and hear their opinions.

Also, @ilicfilip I noticed that there is some code in this pull for the email stuff, should that be in another pull?

ilicfilip:
@jdevalk , my bad - I was working on interactive / email feature and missed to switch branch before making the confetti change. All email code is reverted now.

jdevalk:
Trust me I do this ALL the time

Mijke:
@jdevalk @ilicfilip 

oh I love this confetti!
- maybe add a couple more Ravi's instead of the other badges, for a better ratio between the images 
- maybe it lasts a bit long?? 

and I noticed a couple of small things here:
Scherm­afbeelding 2025-02-24 om 11 11 19


I'll make 2 new issues for other things I noticed

ilicfilip:
@Mijke , I have made the changes to the "upgrade task" popover - here it is how it looks now:
Screenshot 2025-02-24 at 14 00 40

ilicfilip:
Looks good to me, the only thing that I think we can adjust is the size - to me it looks better when objects are a bit smaller, for example setting scalar value to 3.

When heart celebration is triggered, the hearts are now much smaller (they matched the stars size which we used before) so what ever we decide should be applied to them as well.

aristath (approved):

PR #293, opened 2025-02-24
Filip/new page assign term

ilicfilip:
## Context
Fix for the last TODO comment we had. It was about adding the "page types term to the newly created page.

I also fixed CSS issue, check icon was showing if `<select>` option had an option selected (from our "auto detect page" feature), but the `I don't have this page yet` radio was selected.

## Quality assurance

### DIFF
--- a/assets/css/settings-page.css
+++ b/assets/css/settings-page.css
@@ -93,7 +93,7 @@
 			}
 		}
 
-		&:has(option[value=""]:not(:checked)),
+		&:has(option[value=""]:not(:checked)):has(input[type="radio"][value="yes"]:checked),
 		&:has(input[type="radio"][value="not-applicable"]:checked) {
 
 			h3 {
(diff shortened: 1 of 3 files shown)


aristath (changes requested):
Looking at the code, this does not change the option in the UI, does it? 🤔 
We have a page-type selector already in our sidebar (which picks up the assigned term from the taxonomy, so from what I can tell this won't take into account that selector, and even if they select a different option manually in the page-type selector, we'll be overwriting their selection, correct?

Could we instead, assign the term when the editor loads? 🤔 
This way we don't add a custom saving implementation, we just pre-select the right option in our dropdown

ilicfilip:
:+1: I understood differently in Slack discussion.
Changes are reverted and the better way, with our PP selector, is implemented.

I also fixed one small bug with preselected value, when there are no previously assigned terms `select( 'core/editor' ).getEditedPostAttribute( TAXONOMY )` is returning an empty array - which is always true expression. So I corrected check to check for the array length.

PR #294, opened 2025-02-24
Refactor one time classes

jdevalk:
Drastically simplifies how we add providers.


classes/suggested-tasks/local-tasks/providers/class-one-time.php
```diff
@@ -75,4 +205,27 @@ public function get_tasks_to_inject() {
...
+	public function get_task_details( $task_id = '' ) {
+
+		return [
+			'task_id'      => $this->get_task_id(),
+			'title'        => $this->get_title(),
+			'parent'       => $this->get_parent(),
+			'priority'     => $this->get_priority(),
+			'type'         => $this->get_provider_type(),
+			'points'       => 1,
+			'url'          => $this->capability_required() ? \esc_url( $this->get_url() ) : '',
```
jdevalk, line 225: @ilicfilip @aristath is the `capability_required` check here, necessary? We should just not show the task otherwise, right?

classes/suggested-tasks/local-tasks/providers/class-one-time.php
```diff
@@ -75,4 +205,27 @@ public function get_tasks_to_inject() {
...
+	 */
+	public function get_task_details( $task_id = '' ) {
+
+		return [
+			'task_id'      => $this->get_task_id(),
+			'title'        => $this->get_title(),
+			'parent'       => $this->get_parent(),
+			'priority'     => $this->get_priority(),
+			'type'         => $this->get_provider_type(),
+			'points'       => 1,
```
jdevalk, line 224: Should we make this settable too? Actually fairly easy to do now of course.

classes/suggested-tasks/local-tasks/providers/class-one-time.php
```diff
@@ -75,4 +205,27 @@ public function get_tasks_to_inject() {
...
+			'url'          => $this->capability_required() ? \esc_url( $this->get_url() ) : '',
+			'description'  => '<p>' . $this->get_description() . '</p>',
```
jdevalk, line 226: Tempted to move the `<p>` and `</p>` to the `get_description()`, what do you think @aristath @ilicfilip ?

classes/suggested-tasks/local-tasks/providers/class-one-time.php
```diff
@@ -14,6 +14,136 @@
...
+	 * @var int
+	 */
+	protected $parent = 0;
+
+	/**
+	 * The task priority.
+	 *
+	 * @var string
+	 */
+	protected $priority = 'medium';
```
jdevalk, line 64: Not sure if we use this anywhere yet.

ilicfilip: Priority is used in "Suggested tasks" Javascript: https://github.com/ProgressPlanner/progress-planner/blob/d2fdd911e87d0a3316b78d5958e6ac311c3aae5e/assets/js/widgets/suggested-tasks.js#L73-L86

aristath (approved):
This looks good, and simplifies things a lot, I love it!
I only had a small note for which I added a comment, but I'm OK to merge this either way 👍

PR #299, opened 2025-02-25
Fix triggering upgrade popover when it is not needed

ilicfilip:
## Context

I noticed an inconsistent bug, today I managed to replicate it - `progress_planner_previous_version_task_providers` was wrongly overwritten in some cases, which led to upgrade popover being displayed on PP page when it was not needed.
No points were awarded or any side action executed, but it's annoying.

Also a typo was fixed.

## Summary

Fixed unnecessary display of the upgrade popover.

## Quality assurance

### DIFF
--- a/classes/class-plugin-upgrade-handler.php
+++ b/classes/class-plugin-upgrade-handler.php
@@ -34,10 +34,9 @@ class Plugin_Upgrade_Handler {
 	 * @return void
 	 */
 	public function init() {
-
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- We're not processing any data.
 		if ( ! \is_admin() || ( ! isset( $_GET['page'] ) || $_GET['page'] !== 'progress-planner' ) ) {
-			return;
+			return; // Early exit if we're not on the Progress Planner page.
 		}
 
 		// Add the onboarding task providers.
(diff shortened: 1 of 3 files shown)


classes/class-plugin-upgrade-handler.php
```diff
@@ -112,4 +113,18 @@ public function get_newly_added_task_providers() {
...
+	 *
+	 * @return bool
+	 */
+	protected function is_on_progress_planner_page() {
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- We're not processing any data.
+		if ( ! \is_admin() || ( ! isset( $_GET['page'] ) || $_GET['page'] !== 'progress-planner' ) ) {
+			return false;
+		}
+
+		return true;
```
aristath, line 128:
```suggestion
		return \is_admin() && isset( $_GET['page'] ) && $_GET['page'] !== 'progress-planner';
```

classes/class-plugin-upgrade-handler.php
```diff
@@ -112,4 +113,14 @@ public function get_newly_added_task_providers() {
...
 	}
+
+	/**
+	 * Check if we're on the Progress Planner page.
+	 *
+	 * @return bool
+	 */
+	protected function is_on_progress_planner_page() {
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- We're not processing any data.
+		return \is_admin() && isset( $_GET['page'] ) && $_GET['page'] !== 'progress-planner';
```
aristath, line 124:
```suggestion
		return \is_admin() && isset( $_GET['page'] ) && $_GET['page'] === 'progress-planner';
```

aristath (approved):

PR #301, opened 2025-02-26
Add a class to handle plugin migrations 

aristath:
## Context

Adds a new `Plugin_Migrations` class to handle migrating options and data when updating the plugin.

We'll be refactoring and restructuring some data in the suggested-tasks, so we'll need a way to migrate previous data to the new format

### DIFF
--- a/classes/class-plugin-migrations.php
+++ b/classes/class-plugin-migrations.php
@@ -65,7 +65,7 @@ class Plugin_Migrations {
 	 * @return string
 	 */
 	private function get_db_version() {
-		return \get_option( 'progress_planner_version', '1.0.0' );
+		return \get_option( 'progress_planner_version', '1.1.0' );
 	}
 
 	/**
--- a/classes/class-plugin-migrations.php
+++ b/classes/class-plugin-migrations.php
@@ -16,7 +16,7 @@ namespace Progress_Planner;
  *
  * @package Progress_Planner
  */
-class Plugin_Upgrade {
+class Plugin_Migrations {
 
 	/**
 	 * The plugin database version.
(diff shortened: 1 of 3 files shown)


classes/class-plugin-migrations.php
```diff
@@ -0,0 +1,102 @@
...
+		return \get_file_data( PROGRESS_PLANNER_FILE, [ 'Version' => 'Version' ] )['Version'];
+	}
+
+	/**
+	 * Get the plugin database version.
+	 *
+	 * @return string
+	 */
+	private function get_db_version() {
+		return \get_option( 'progress_planner_version', '1.0.0' );
```
ilicfilip, line 68: Should default version here be 1.1.0?

classes/class-plugin-migrations.php
```diff
@@ -0,0 +1,102 @@
...
+		foreach ( self::UPGRADE_METHODS as $version => $upgrade_method ) {
+			if (
+				( defined( 'PRPL_DEBUG' ) && PRPL_DEBUG ) ||
+				\get_option( 'prpl_debug' ) || version_compare( $version, $this->db_version, '>' )
+			) {
+				$this->$upgrade_method();
+			}
+		}
+
+		\update_option( 'progress_planner_version', $this->version );
```
ilicfilip, line 92: Does it make sense to add a `do_action` here?

For example, in [Plugin_Update_Handler](https://github.com/ProgressPlanner/progress-planner/blob/1e58bc2685bfca6b6ef96a43050a13151b9e51c5/classes/class-plugin-upgrade-handler.php) (which will now be renamed) there is a code which needs to be run on every update.

So instead of adding `upgrade_method` in this class for every update it would be possible to hook to this new action?
(There is a WP hook when plugin is updated, we just need to check if it is PP before running the code - but this would be more aligned with the our new upgrade code)

aristath: The problem is that a hook like that would not work when updating the plugin via FTP etc, so I deemed it safer to do it like this...

ilicfilip (approved):

PR #303, opened 2025-02-26
Tweak upsell challenges

ilicfilip:
## Context
Change the styling of the Challenges widget.

## Quality assurance


aristath:
@ilicfilip please keep in mind that these should be replicable when we change the content server-side, without needing to update the plugin, and some challenges will also be available to free users

aristath (approved):

PR #306, opened 2025-02-27
Simplify suggested tasks script

aristath:
## Context
This PR simplifies the suggested-tasks script:
Merges the task details in the tasks list, and adds a `status` key in them.
This will allow us to more efficiently refactor the options and how they are saved

### DIFF
--- a/classes/widgets/class-suggested-tasks.php
+++ b/classes/widgets/class-suggested-tasks.php
@@ -141,7 +141,12 @@ final class Suggested_Tasks extends Widget {
 					'high'   => 0,
 					'medium' => 1,
 					'low'    => 2,
+					'none'   => 3,
 				];
+
+				$a['priority'] = ! isset( $a['priority'] ) || ! isset( $priority[ $a['priority'] ] ) ? 'none' : $a['priority'];
+				$b['priority'] = ! isset( $b['priority'] ) || ! isset( $priority[ $b['priority'] ] ) ? 'none' : $b['priority'];
+
 				return $priority[ $a['priority'] ] - $priority[ $b['priority'] ];
 			}
 		);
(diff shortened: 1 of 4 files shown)


classes/widgets/class-suggested-tasks.php
```diff
@@ -114,33 +114,60 @@ public function enqueue_scripts() {
...
+		$final_tasks = array_values( $final_tasks );
+
+		// Sort the final tasks by priority. The priotity can be "high", "medium", "low", or "none".
+		uasort(
+			$final_tasks,
+			function ( $a, $b ) {
+				$priority = [
+					'high'   => 0,
+					'medium' => 1,
+					'low'    => 2,
```
ilicfilip, line 143: Can we add here a check if `$priority[ $key ] ` is set (and maybe set to a default value if not)?

For example, if I set priority to `none` in task details a PHP warning is triggered:
`PHP Warning:  Undefined array key "none"`

aristath: Good catch, should be fixed now 👍

ilicfilip:
There seems to be a bug when Task is dismissed - I tried it on review post task but it is the same for Remote tasks
Screenshot 2025-02-27 at 08 51 02

aristath:
Fixed 👍

ilicfilip:
One more, for some reason "Perform all updates" is not removed from the widget when it is snoozed (it is after page reload):

https://github.com/user-attachments/assets/abe7508c-7b92-4be5-9fd8-178c07de34d2

aristath:
Thanks @ilicfilip! It was a typo, fixed ❤️

ilicfilip (approved):

PR #307, opened 2025-02-27
Simplify suggested-tasks script dependencies

aristath:
We perform many unnecessary calculations in order to determine if the confetti script should be loaded or not.
This PR removes all those and always loads the confetti script to simplify the implementation.


ilicfilip (approved):
The reason was to prevent enqueueing script(s) on the WP Dashboard screen (since we show the widget there as well) - but I agree, it became too complex 👍

PR #310, opened 2025-02-27
Promisify progressPlannerAjaxRequest

ilicfilip:
## Context
This PR changes how we make AJAX requests, until now along with the `data` we passed both `Success` and `Fail` callbacks to the `progressPlannerAjaxRequest` function.

Although this was working, it was a bit involved to make changes since sometimes we have nested AJAX requests and making an action when we need to wait for multiple AJAX requests to be done was a lot more complicated then it should be.

Now we can simply do:

```
progressPlannerAjaxRequest( {
  url: progressPlanner.onboardAPIUrl,
  data,
} )
.then( ( apiResponse ) => {
// 
} )
.catch( error ) => {
//
} );
```

Example of executing code when multiple AJAX requests are done is:

```
// Start scanning posts.
const scanPromise = progressPlannerTriggerScan();

// Start the tasks.
const tasksPromise = prplOnboardTasks();

// Wait for all promises to resolve.
Promise.all( [ scanPromise, tasksPromise ] ).then( () => {
  // All promises resolved, redirect to the next step.
  prplOnboardRedirect();
} );
```

## Quality assurance

### DIFF
--- a/assets/js/ajax-request.js
+++ b/assets/js/ajax-request.js
@@ -16,45 +16,51 @@ const progressPlannerAjaxRequest = ( {
 	successAction,
 	failAction,
 } ) => {
-	const http = new XMLHttpRequest();
-	http.open( 'POST', url, true );
-	http.onreadystatechange = () => {
-		const defaultCallback = ( response ) => {
-			// eslint-disable-next-line no-console
-			console.info( response );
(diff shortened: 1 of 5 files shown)


aristath (approved):
Appears to be working, but I did not test onboarding/posts-scanning

PR #311, opened 2025-02-28
WIP - Migrate the `progress_planner_suggested_tasks` option

aristath:
## Context

*

## Summary

*

*

#### Relevant test scenarios
* [ ] Changes should be tested with the browser console open
* [ ] Changes should be tested on different posts/pages/taxonomies/custom post types/custom taxonomies
* [ ] Changes should be tested on different editors (Block/Classic/Elementor/other)
* [ ] Changes should be tested on different browsers
* [ ] Changes should be tested on multisite

## UI changes

* [ ] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

## Documentation

## Quality assurance

Fixes #

### DIFF
--- a/classes/class-suggested-tasks.php
+++ b/classes/class-suggested-tasks.php
@@ -32,13 +32,6 @@ class Suggested_Tasks {
 	private $remote;
 
 	/**
-	 * The name of the settings option.
-	 *
-	 * @var string
-	 */
-	const OPTION_NAME = 'progress_planner_suggested_tasks';
-
-	/**
 	 * Constructor.
 	 *
 	 * @return void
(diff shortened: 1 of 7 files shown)


classes/class-suggested-tasks.php
```diff
@@ -435,17 +356,23 @@ public function snooze_task( $task_id, $duration ) {
...
+		$tasks_changed = false;
+		foreach ( $tasks as $key => $task ) {
+			if ( $task['status'] !== 'snoozed' ) {
+				continue;
+			}
 
-		foreach ( $option['snoozed'] as $task ) {
-			if ( $task['time'] < $current_time ) {
-				$this->remove_task_from( 'snoozed', $task['id'] );
+			if ( isset( $task['time'] ) && $task['time'] < \time() ) {
```
ilicfilip, line 366: I think here we should just remove task, instead of setting it to pending.

We did it until now and I think the reason was that by the time the snooze period passes task my not be relevant (ie `update-core-202451`, we have a cleanup method for "outdated" tasks like that).

Also all available task providers are checked on every page load, so task will be added back in any case.

ilicfilip (changes requested):

ilicfilip:
I have tested a bit, debug tools needed to be migrated and task celebration wasn't working so I ported the code we had before.

Other than that, and the comment I left above, everything seems to be working fine.

classes/class-suggested-tasks.php
aristath: After this PR I am planning to migrate TODO items and combine them with the suggested tasks... So basically, they will be suggested items under a new provider `user`.
With that in mind, do you think we should delete them (and add a check to bypass `user` tasks? Or should we keep it as-is and just change the status? 🤔

aristath: Thanks! Pushed a tweak for this in https://github.com/ProgressPlanner/progress-planner/pull/311/commits/bd45f498010ac98386c85ef0dc14d4bc06650add 👍

aristath:
@ilicfilip are we OK to merge this?

ilicfilip (approved):

ilicfilip:
Approved 👍

PR #319, opened 2025-03-04
migrate user todos

aristath:
This PR migrates the internal structure of user TODO items:
Instead of having a separate implementation for saving, retrieving and managing TODO items, we are now saving them as suggested tasks by adding a new `user` provider.


classes/class-rest-api-stats.php
aristath, line 1: Note: Ravi's tasks should be added here.
How the purposes of the POC I have completely removed the todos from the REST-API
Once that's done, we'll also need to refactor the server-side implementation as well to accommodate the new data.

tacoverdo:
Something seems to be going wrong with completed tasks and moving them to 'completed tasks'.
CleanShot 2025-03-18 at 10 03 25

And we have some interesting behaviour when checking and unchecking Ravi's Recommendations:
CleanShot 2025-03-18 at 09 57 54

Mijke:
@tacoverdo @aristath I think it will make more sense for me to review when [the design / styling](https://github.com/ProgressPlanner/progress-planner-pro/issues/72#issuecomment-2732917329) I have to make is implemented

aristath:
@Mijke of course! If you want to check where we are right now, you can use the playground link on the [1st comment](https://github.com/ProgressPlanner/progress-planner/pull/319#issuecomment-2696654173) 👍 (implemented the checkboxes as agreed in our previous discussion)

ilicfilip (approved):

PR #320, opened 2025-03-04
Convert "review content" to repetitive task 

ilicfilip:


aristath:
@ilicfilip is this one just a comment? 🤔

ilicfilip:
It was https://github.com/ProgressPlanner/progress-planner/pull/320/commits/929f1d63e3b4d626d14f7ccdbfe5c75679c5ea51, but we misunderstood - so I reverted that commit.

I plan to do this, but I need to wrap up https://github.com/ProgressPlanner/progress-planner/pull/316 first.

classes/suggested-tasks/local-tasks/class-local-task-factory.php
```diff
@@ -13,19 +13,19 @@
...
 
 	/**
-	 * The task ID.
+	 * The task ID or task data.
 	 *
-	 * @var string
+	 * @var mixed
 	 */
-	private $task_id;
+	private $task;
```
aristath, line 20: I don't particularly like this...
A var that could be either a task-ID or a task-data array makes things more complicated for the future since it's not clear what we have and we'll need to add additional checks etc in the future if we want to do other things.
Could we make sure it's ALWAYS the data and not an ID? 🤔

ilicfilip: I did it to spare us doing the following in our codebase (as that will be frequent now):
```
$data = \progress_planner()->get_suggested_tasks()->get_task_by_task_id( $this->task );
$task_object = ( new Local_Task_Factory( $task_data ) )->get_task()
```

I dont think we create task from data yet anywhere, but we can change the implementation so we can use:
```
Local_Task_Factory::create_task_from_id( $task_id );
Local_Task_Factory::create_task_from_data( $task_data );
```

?

aristath:
```php
Local_Task_Factory::create_task_from( 'id', $task_id );
Local_Task_Factory::create_task_from( 'data', $task_data );
```
?

ilicfilip: Done

aristath (approved):

PR #325, opened 2025-03-05
Refactor/rename types and ids

aristath:
## Context
This PR mostly renames some things and does minor tweaks.
Suggested tasks had some confusing names for vars, constants and methods:
* `type` is a very vague term... in some cases was the status, in other cases it was the category, and in other cases it was the provider ID. This PR differentiates between the three, renaming `TYPE` to `CATEGORY` and then going through the code to properly use `status`, `category`, and `provider_id` where we used `type`.
* `id` in some cases was the provider ID, in other cases it was the task ID. These two are sometimes the same, but not always. This PR renamed the `ID` in providers to `PROVIDER_ID` and makes a distinction between the two.

### DIFF
--- a/classes/class-plugin-migrations.php
+++ b/classes/class-plugin-migrations.php
@@ -159,6 +159,15 @@ class Plugin_Migrations {
 			}
 		}
 
+		$convert_task_id = function ( $task_id ) {
+			if ( ! str_contains( $task_id, '|' ) ) {
+				return $task_id;
+			}
+			$parts = \explode( '|', $task_id );
+			\ksort( $parts );
+			return \implode( '|', $parts );
+		};
+
 		foreach ( $local_tasks as $key => $task ) {
 			if ( isset( $task['type'] ) ) {
 				$local_tasks[ $key ]['category'] = $task['type'];
@@ -166,6 +175,7 @@ class Plugin_Migrations {
 				unset( $local_tasks[ $key ]['type'] );
 				$local_tasks_changed = true;
 			}
+			$local_tasks[ $key ]['task_id'] = $convert_task_id( $task['task_id'] );
 		}
 
 		if ( $local_tasks_changed ) {
@@ -173,14 +183,13 @@ class Plugin_Migrations {
 		}
 
 		// Migrate acgtivities saved in the progress_planner_activities table.
-		$activities = \progress_planner()->get_query()->query_activities(
+		foreach ( \progress_planner()->get_query()->query_activities(
 			[ 'category' => 'suggested_task' ],
-		);
-
-		foreach ( $activities as $activity ) {
-			$data_id = $activity->data_id;
-			if ( str_contains( $data_id, 'type' ) ) {
-				$activity->data_id = str_replace( 'type', 'category', $data_id );
+		) as $activity ) {
+			$data_id     = $activity->data_id;
+			$new_data_id = $convert_task_id( $data_id );
+			if ( $new_data_id !== $data_id ) {
+				$activity->data_id = $new_data_id;
 				$activity->save();
 			}
 		}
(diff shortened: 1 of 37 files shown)


ilicfilip:
Looks like the checks which include `prplSuggestedTasks.maxItemsPerProviderID` are not working correctly, I see all pending tasks in the widget (but the `prplSuggestedTasks.maxItemsPerProviderID` values are correct)

classes/suggested-tasks/class-remote-tasks.php
```diff
@@ -46,10 +46,10 @@ public function inject_tasks( $tasks ) {
 				continue;
 			}
 
-			// Task which don't have type defined are added as a 'default' type.
-			$item['type']    = 'remote-' . ( isset( $item['type'] ) ? $item['type'] : 'default' );
-			$item['task_id'] = "remote-task-{$item['task_id']}";
-			$items[]         = $item;
+			// Task which don't have category defined are added as a 'default' category.
+			$item['category'] = 'remote-' . ( isset( $item['category'] ) ? $item['category'] : 'default' );
```
ilicfilip, line 50: This change also needs to be applied on the server side, [here](https://github.com/ProgressPlanner/progress-planner-server/blob/ef672783acd13af41256fffd6428f584fa8cdd70/includes/rest-api/class-get-suggested-tasks.php#L109)

classes/class-plugin-migrations.php
```diff
@@ -159,6 +159,13 @@ private function upgrade_1_1_1() {
 			}
 		}
 
+		foreach ( $local_tasks as $key => $task ) {
+			if ( isset( $task['type'] ) ) {
+				$local_tasks[ $key ]['category'] = $task['type'];
```
ilicfilip, line 164: Here we should update the `task_id` as well, for example `date/202510|long/1|type/create-post` - if not it results in duplicated tasks with the Pending status, one with new and one with old format (since we compare if task is already added by comparing the `task_id`).

<details><summary>Screenshot</summary>
Screenshot at Mar 05 09-30-14
</details>


If task is changed then probably `$local_tasks_changed = true;` should be added.

I think we should migrate the Snoozed and Pending celebration tasks, since that will not work correctly with old format (when checking if task is snoozed we also checked for task `type` / now `category`). If we're doing it all we can then migrate the Completed tasks as well.

On a side note, we also have task_ids saved in the `wp_progress_planner_activities` table (`data_id` column). Since we use it to calculate points for monthly badges probably we should migrate to new format.

aristath:
> Looks like the checks which include `prplSuggestedTasks.maxItemsPerProviderID` are not working correctly, I see all pending tasks in the widget (but the `prplSuggestedTasks.maxItemsPerProviderID` values are correct)

Yeah, I was counting items per provider-ID while it should be per category. Pushed a fix 👍

classes/class-plugin-migrations.php
ilicfilip: yes for the Local tasks.. but should we do it for the activities? I am not sure that calcs will work (for the badges which are not completed) if task_id is not changed there as well.

ilicfilip: I pushed this one to fix wrong array key: https://github.com/ProgressPlanner/progress-planner/pull/325/commits/aa4ec0b68a68f4c9f88b8164e467c825ed0895ff

ilicfilip: But that didn't fix the issue, there is still mismatch with the task_ids:

- migrated: `date/202510|long/1|category/create-post`
- newly created: `category/create-post|date/202510|long/1`

When creating new task we also do a ksort [here](https://github.com/ProgressPlanner/progress-planner/blob/aa4ec0b68a68f4c9f88b8164e467c825ed0895ff/classes/suggested-tasks/local-tasks/class-local-task-factory.php#L83), which moved `type` to the end of the array but now it sets `category` as the beginning.

So probably we have to do the same when migrating.

ilicfilip:
> I believe this was fixed now in https://github.com/ProgressPlanner/progress-planner/commit/67451cf2c40e90abf37218fef4a9cadfca8d54cb

For me it still wasn't working, I opened a PR for it: https://github.com/ProgressPlanner/progress-planner/pull/328

ilicfilip (approved):

PR #332, opened 2025-03-06
Tweaks for injecting pending-celebration tasks

aristath:


ilicfilip:
I have found a bug, `pending_celebration` posts are injected multiple times:
Screenshot 2025-03-06 at 12 33 34

I think the reason is that now `pending_celebration` tasks have their category unchanged, they are both injected [here](https://github.com/ProgressPlanner/progress-planner/blob/8c6fc0339bde12f22edfe1f9099180927b9acd2b/assets/js/widgets/suggested-tasks.js#L298) and [here](https://github.com/ProgressPlanner/progress-planner/blob/8c6fc0339bde12f22edfe1f9099180927b9acd2b/assets/js/widgets/suggested-tasks.js#L304).

aristath:
@ilicfilip good catch! Should be OK now, could you please confirm?

ilicfilip (approved):

aristath:
🤦 You're right.. Thanks for fixing it!

PR #338, opened 2025-03-07
Add init method to Task Providers

ilicfilip:
## Context

While working on other feature I noticed that we did the `( new Review() )` in our code, purpose was just to get the provider ID (so it's not hardcoded). This is in general what we want, but.. that class adds hooks in it's constructor, so now they were added twice.

This was still a WIP code, and I dont think it is correct since we dont need to check if edited post was long or short (this condition is changed in https://github.com/ProgressPlanner/progress-planner/pull/316, which will hopefully be merged soon) but it got me worried since this is easy to overlook.

For this to not happen again I have added the init method, which is called only once (when task provider is added to the local task manager).

## Quality assurance

### DIFF
--- a/classes/activities/class-suggested-task.php
+++ b/classes/activities/class-suggested-task.php
@@ -8,7 +8,7 @@
 namespace Progress_Planner\Activities;
 
 use Progress_Planner\Activity;
-use Progress_Planner\Suggested_Tasks\Local_Tasks\Providers\Content\Review;
+use Progress_Planner\Suggested_Tasks\Local_Tasks\Providers\Content\Create;
 
 /**
  * Handler for suggested tasks activities.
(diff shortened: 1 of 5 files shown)


classes/activities/class-suggested-task.php
```diff
@@ -74,7 +74,7 @@ public function get_points( $date ) {
 			true === $data['long'] &&
 			(
 				'create-post' === $data['provider_id'] ||
-				( new Review() )->get_provider_id() === $data['provider_id']
+				'review-post' === $data['provider_id']
```
aristath, line 77: I agree with adding an `init` method, but if we're adding that method, then why change this one? 🤔

ilicfilip: basically this:
> and I dont think it is correct since we dont need to check if edited post was long or short (this condition is changed in https://github.com/ProgressPlanner/progress-planner/pull/316, which will hopefully be merged soon) but it got me worried since this is easy to overlook.

I found another instance of it, [done for the same purpose](https://github.com/ProgressPlanner/progress-planner/blob/7255b21682aff21933bf2b8fc4e3d3b7862daa42/classes/widgets/class-suggested-tasks.php#L146), and I left since we need that one.

aristath (approved):

PR #343, opened 2025-03-11
Reduce number of database queries when checking for user activity scores

ilicfilip:
## Context
While inspecting performance I noticed unusual number of database queries we make, even more strange was that a dates from the past were used in them.

It turns out we query for activities 2 years in the past, regardless of when the plugin was activated.

This PR changes that, so we make queries from the activation date or 2 years ago (what ever is more recent).
For me it reduced number of queries from 217 to 111.

## Summary

Reduced number of database queries when checking for user activity scores.

## Quality assurance

### DIFF
--- a/classes/widgets/class-activity-scores.php
+++ b/classes/widgets/class-activity-scores.php
@@ -166,6 +166,10 @@ final class Activity_Scores extends Widget {
 	 * @return array
 	 */
 	public function personal_record_callback() {
+		$activation_date = \progress_planner()->get_base()->get_activation_date();
+		$two_years_ago   = new \DateTime( '-2 years' );
+		$start_date      = $activation_date < $two_years_ago ? $two_years_ago : $activation_date;  // If the activation date is more than 2 years ago, set it to 2 years ago.
+
 		$goal = Goal_Recurring::get_instance(
 			'weekly_post_record',
 			[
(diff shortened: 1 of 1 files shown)


aristath:
The record is supposed to be an all-time record, so limiting it to 2 years doesn't feel right...
Can we find another way to reduce the number of these queries?
Caching them should be more effective without changing the logic... 🤔

aristath:
I think using the existing `progress_planner()->get_settings()` implementation would be preferable...
Transients have issues on some hosts, so it would make sense to just add a new entry in our main option to store previous months data, should be a minimal amount of data anyway.
Like you said we rarely (if ever) delete previous month activities, so it makes sense to just store it there 👍

classes/widgets/class-activity-scores.php
```diff
@@ -180,26 +183,52 @@ public function personal_record_callback() {
...
+					\progress_planner()->get_settings()->set( $this->cache_key, $cached_activities );
+
+					// Return the cached value.
+					return $cached_activities[ $weekly_cache_key ];
 				},
 			],
 			[
 				'frequency'     => 'weekly',
-				'start_date'    => $start_date,
+				'start_date'    => \progress_planner()->get_base()->get_activation_date(),
```
aristath, line 217: I don't think `get_base()` is necessary, `\progress_planner()` is already an instance of the `Base` class

ilicfilip: Done.

aristath (approved):

PR #348, opened 2025-03-14
E2E testing

ilicfilip:


classes/class-rest-api-tasks.php
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	public function register_rest_endpoint() {
+		\register_rest_route(
+			'progress-planner/v1',
+			'/tasks',
+			[
+				[
+					'methods'             => 'GET',
+					'callback'            => [ $this, 'get_tasks' ],
+					'permission_callback' => '__return_true',
```
aristath, line 37: We should move the authentication you now have in the `get_tasks` method to a new method and use it as a `permission_callback` 👍

ilicfilip: Done.

ilicfilip:
This PR adds e2e (Playwright) tests to our CI. 
Currently following following checks are made:

- It is verified that Progress Planner Dashboard is loaded and that Onboard is working (without sharing email)
- Checks if the Tour is started when it is clicked on the button in the header section (Lightbulb icon)
- It is checked that Post Types setting is properly saved after change (Settings icon)
- If we have a dismissible task, it dismisses it and verifies that is marked as completed in the database
- Completes the "Set tagline" task and verifies that is celebrated and marked as completed in the database
- Picks one other task and snoozes it, again verifies that is actually marked as snoozed in the database
- It creates and deletes a todo item to verify it works properly (refreshes the page, to make sure it's not just added / removed from the DOM)

aristath (approved):

PR #360, opened 2025-03-21
Refactor stylesheets enqueueing

aristath:
Similar to #353, this PR refactors the way we enqueue stylesheets.

### DIFF
--- a/classes/admin/class-dashboard-widget-score.php
+++ b/classes/admin/class-dashboard-widget-score.php
@@ -46,12 +46,7 @@ class Dashboard_Widget_Score extends Dashboard_Widget {
 			$suggested_tasks_widget->enqueue_scripts();
 		}
 
-		\wp_enqueue_style(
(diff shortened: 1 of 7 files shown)


ilicfilip:
While reviewing this PR I found a potential bug which is affecting the new enqueueing system in general, infinite recursion will occur if scripts reference each other:

For example in `assets/js/widgets/suggested-tasks.js` we have:
`Dependencies: progress-planner/grid-masonry, progress-planner/web-components/prpl-suggested-task, progress-planner/document-ready, particles-confetti`

And we add it as a dependency in one of those scripts, for example in `assets/js/grid-masonry.js`:
`Dependencies: progress-planner/document-ready, progress-planner/widgets/suggested-tasks`

The infinite recursion will occur.

I tried to do an early exit, but it wasn't enough as the recursion was actually occurring within the WP core.

@aristath , if you think this case is not important you can revert the last commit.

ilicfilip (approved):

PR #365, opened 2025-03-21
Simplify tooltip implementation

ilicfilip:
## Context

Currently we have only 3 tooltips, but implementation is basic and most of the code is repeated (opening, closing and etc).

For the upcoming release we need to add at least one more tooltip and add an overlay when tooltips are opened - so this is a good opportunity to refactor it (as much as the timeframe allows).

## Quality assurance


ilicfilip:
Tooltip component is implemented now, a few notes:
- Overlay is triggered on PP Dashboard page but not on WP Dashboard page, to me it felt weird to do it there. If we want it then just [this](https://github.com/ProgressPlanner/progress-planner/pull/365/files#diff-87edb41eb05dc92769f6f9e142e85de74d524c204d7f1e4a6746540b34d19033R39) element.
- Tooltip and checkbox styling for the RR which can't be dismissed by the user are added based on [this comment](https://github.com/ProgressPlanner/progress-planner-pro/issues/96#issuecomment-2747889530)

assets/css/admin.css
aristath, line 1: We now have an `assets/css/web-components` folder, perhaps the styles for the new web-component could be moved there?

ilicfilip: Styles are moved to it's own file.

The file is enqueued always on our pages since it is used by multiple widgets (and I am about to add new tooltip when this one is merged).

PR #371, opened 2025-03-25
Tweak Golden task implementation

ilicfilip:
## Context

While testing the new User / Todo tasks I encountered a few bugs, at least they looked like bugs to me, for which it was better to open PR.

The most obvious one was that Golden task wasn't set, unless there is at least one completed Todo task:
<details><summary>
Screen recording
</summary>
https://github.com/user-attachments/assets/12cfdadb-2cc0-451c-b7a0-e60093fcff9c
</details>

I believe [this check](https://github.com/ProgressPlanner/progress-planner/blob/ari/migrate-user-todos/classes/class-todo.php#L173-L175) was the problem, since it will `return` if there are no completed activities (which belong to the Todo tasks).

Another issue, which I haven't recorded, was that sometimes completed task was set as the Golden task. For that one I think the problem was that we set the first element of $items array as golden [here](https://github.com/ProgressPlanner/progress-planner/blob/ari/migrate-user-todos/classes/class-todo.php#L193), but `get_items` method returns all tasks (not just pending).

This PR does the following:

- When task is added it checks if it needs to be set as golden or not. There is only 1 golden per week, so if Golden task is deleted (still needs to be decided if this should be allowed or not) user can create new one which will be set as Golden. This is done without as soon as task is added (without refreshing the page).
- Golden task is always added at the top of the Todo list, user can move it but after the page refresh it will again be at the top (should we hide the up / down arrows?)
- Since we don't use order for completed tasks, order is removed from the task when it is marked as completed. If task is again set as pending it will be added to the end of the Todo list (internally we set `max order + 1` for tasks which don't have order set). It was done to prevent multiple tasks having the same order.
- A check is added which runs on Monday (or at the first occasion user uses his website after the Monday) which checks which task is at the top of the list and sets it as Golden.
- Descriptions for Golden and Silver tasks are added per [this comment](https://github.com/ProgressPlanner/progress-planner-pro/issues/96#issuecomment-2740427585). If [this PR ](https://github.com/ProgressPlanner/progress-planner/pull/365)is approved we can add tooltips easily as well (it will be much easier then implementing custom tooltip for this purpose).

## Quality assurance

### DIFF
--- a/assets/css/page-widgets/todo.css
+++ b/assets/css/page-widgets/todo.css
@@ -25,6 +25,33 @@
 			height: 1.25em;
 		}
 	}
+
+	.prpl-todo-golden-task-description,
+	.prpl-todo-silver-task-description {
+		display: none;
+	}
+
+	&:not(:has(#todo-list li[data-task-points="1"])) {
+
+		.prpl-todo-golden-task-description {
+			display: none;
+		}
+
+		.prpl-todo-silver-task-description {
+			display: block;
+		}
+	}
+
+	&:has(#todo-list li[data-task-points="1"]) {
+
+		.prpl-todo-silver-task-description {
+			display: none;
+		}
+
+		.prpl-todo-golden-task-description {
+			display: block;
+		}
+	}
 }
 
 #create-todo-item {
--- a/assets/js/widgets/todo.js
+++ b/assets/js/widgets/todo.js
@@ -106,28 +106,36 @@ prplDocumentReady( () => {
 				order: prplGetHighestTodoItemOrder() + 1,
 			};
 
-			// Inject the new task into the DOM.
-			document.dispatchEvent(
-				new CustomEvent( 'prpl/todo/injectItem', {
-					detail: {
-						item: newTask,
-						addToStart: false,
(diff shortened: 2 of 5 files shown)


assets/css/page-widgets/todo.css
```diff
@@ -25,6 +25,51 @@
...
+		}
+
+		#todo-list > *:first-child {
+
+			.prpl-suggested-task {
+				border: 1px solid var(--prpl-color-accent-orange);
+				background-color: var(--prpl-background-orange);
+			}
+		}
+	}
```
aristath, line 72: This doesn't work properly... If the user reorders the list and the golden task is no longer at the top of the list, then the border & background are applied to the 1st item on the list instead of the "golden" task 🤔

ilicfilip: Yes, we have an open question about disabling up / down controls for the golden task since after the reload golden task will again be at the top (and that can be confusing for the user).

I will change styling now so it is applied only to the +1 task (no matter where it is positioned).

aristath (approved):

PR #373, opened 2025-03-25
Once permalink task is completed show icon only next to selected option

ilicfilip:
## Context

Idea from @tacoverdo , discussed on Slack.

For "Set permalink" task we show Ravi icon next to 2 options which we recommend for user to select. But after one of them was selected (and permalink options saved) we should show only icon next to the selected one (not both).

## Quality assurance


tacoverdo:
CleanShot 2025-03-25 at 15 56 43

I think I found a bug. After selecting a recommended setting, the icon changes to the check icon and the secondary icon disappears. However, when then selecting a not-recommended setting, the icons still show the green check version, instead of the +1 for the recommended settings.

aristath (approved):

Issue #372, opened 2025-03-25
Multi line tasks pushing "add new task" input outside the container


goodplansam:
* [x] I've read and understood the [contribution guidelines](https://github.com/Emilia-Capital/.github/blob/main/CONTRIBUTING.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Step-by-step reproduction instructions
1. Writing in the task, hitting enter, drops a new line in the same task.
2. Writing another line of text in the same task. 
3. Repeating step two a few times. 

#### Expected results
1. On hitting enter should refocus onto the "add new task" input field. 

#### Actual results
1. "Add a new task"  input drops out of the container. 

### Screenshots, screen recording, code snippet
If possible, please provide a screenshot, a screen recording or a code snippet which demonstrates the bug.

![Image](https://github.com/user-attachments/assets/ea5c1579-f62c-47da-8d16-e83acb98bcf9)

### Technical info

* If relevant, which editor is affected (or editors): 
- [ ] Block Editor
- [ ] Gutenberg Editor
- [ ] Elementor Editor
- [ ] Classic Editor 
- [ ] Other: 

* Which browser is affected (or browsers): 
- [*] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Other: 

#### Used versions
* Device you are using: pc/windows
* Operating system: win 11
* PHP version: 
* WordPress version: 
* WordPress Theme:
* Plugin version: 
* Gutenberg plugin version: 
* Elementor plugin version: 
* Classic Editor plugin version: 
* Relevant plugins in case of a bug: 

Using this playground link: https://playground.wordpress.net/#%7B%22landingPage%22:%22/wp-admin/admin.php?page=progress-planner%22,%22features%22:%7B%22networking%22:true%7D,%22login%22:true,%22plugins%22:%5B%22https://github-proxy.com/proxy/?repo=ProgressPlanner/progress-planner&branch=ari/migrate-user-todos%22%5D,%22steps%22:%5B%7B%22step%22:%22defineWpConfigConsts%22,%22consts%22:%7B%22IS_PLAYGROUND_PREVIEW%22:true%7D%7D%5D%7D

goodplansam:
Oh when I have deleted the multiple lines of text the container is dose not resize to the size of the actual todo list 

![Image](https://github.com/user-attachments/assets/49398cfa-1b04-4ed3-a766-233fc3409c70)

ilicfilip:
[...]

PR #376, opened 2025-03-26
Prevent user adding a new line when editing TODO task title

ilicfilip:
## Context

This PR prevents user from inserting line breaks when TODO task title is edited (which is not allowed when todo item is created).

Maybe a bit of an edge case, but user can still enter multiple lines title which can push "Add new task" input out of the widget.

@aristath , I don't know how costly is to re-size grid on every keydown event, but if it is not maybe we trigger that event?

## Quality assurance

Fixes https://github.com/ProgressPlanner/progress-planner/issues/372

### DIFF
--- a/assets/js/web-components/prpl-suggested-task.js
+++ b/assets/js/web-components/prpl-suggested-task.js
@@ -441,32 +441,48 @@ customElements.define(
 
 			// When an item's contenteditable element is edited,
 			// save the new content to the database
-			this.querySelector( 'h3 span' ).addEventListener( 'input', () => {
-				// Add debounce to the input event.
-				clearTimeout( this.debounceTimeout );
-				this.debounceTimeout = setTimeout( () => {
-					const title = this.querySelector( 'h3 span' ).textContent;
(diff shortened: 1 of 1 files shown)


aristath:
> @aristath , I don't know how costly is to re-size grid on every keydown event, but if it is not maybe we trigger that event?

It is costly...

aristath:
Instead of changing the current implementation, could we not just add this?
```js
this.querySelector( 'h3 span' ).addEventListener( 'keydown', ( e ) => {
	if ( e.keyCode === 13 ) {
		e.preventDefault();
	}
} );
```
this _should_ prevent newlines, without affecting how other things work.

ilicfilip:
I have looked at it, but 2 things made me implement different solution:
- It needs to cover both Enter and Return, is 13 keycode for both (I can't test on Windows)?
- It is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)

aristath (approved):

PR #380, opened 2025-03-26
Compatibility tweak for SQLite

aristath:
SQLite doesn't support `MODIFY`.
This PR changes the db upgrade routine to add compatibility for that scenario


jdevalk:
LGTM, so let's fix like this for now. 

The bigger question I have is whether we shouldn't be fixing this in the SQLite integration instead of here...

aristath:
> The bigger question I have is whether we shouldn't be fixing this in the SQLite integration instead of here...

Yes, we should. The problem is that I don't know how... In the past few months they have completely changed the parser, and I have no idea what to make of it, I haven't had time to dive into it yet 😅

PR #389, opened 2025-03-28
Move classes around to improve structure

aristath:
What it says in the title.

This PR moves classes in subfolders to improve our structure.
In addition, we also add backward-compatibility tweaks, both for methods and classes.


ilicfilip (approved):
Maybe we merge this one after the release?

PR #398, opened 2025-04-01
Delay creating local task objects

ilicfilip:
## Context

There are PHP warnings triggered when core is updated to 6.8 RC1:
<details><summary>error_log</summary>
```
[01-Apr-2025 14:35:42 UTC] PHP Notice:  Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>progress-planner</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.) in /Users/filip/Valet/planner/htdocs/wp-includes/functions.php on line 6121
[01-Apr-2025 14:35:42 UTC] PHP Stack trace:
[01-Apr-2025 14:35:42 UTC] PHP   1. {main}() /Users/filip/.composer/vendor/laravel/valet/server.php:0
[01-Apr-2025 14:35:42 UTC] PHP   2. require() /Users/filip/.composer/vendor/laravel/valet/server.php:110
[01-Apr-2025 14:35:42 UTC] PHP   3. require_once() /Users/filip/Valet/planner/htdocs/wp-admin/admin.php:35
[01-Apr-2025 14:35:42 UTC] PHP   4. require_once() /Users/filip/Valet/planner/htdocs/wp-load.php:50
[01-Apr-2025 14:35:42 UTC] PHP   5. require_once() /Users/filip/Valet/planner/htdocs/wp-config.php:112
[01-Apr-2025 14:35:42 UTC] PHP   6. include_once() /Users/filip/Valet/planner/htdocs/wp-settings.php:545
[01-Apr-2025 14:35:42 UTC] PHP   7. progress_planner() /Users/filip/Valet/planner/htdocs/wp-content/plugins/progress-planner/progress-planner.php:70
[01-Apr-2025 14:35:42 UTC] PHP   8. Progress_Planner\Base->init() /Users/filip/Valet/planner/htdocs/wp-content/plugins/progress-planner/progress-planner.php:66
[01-Apr-2025 14:35:42 UTC] PHP   9. Progress_Planner\Base->__call($name = 'get_todo', $arguments = []) /Users/filip/Valet/planner/htdocs/wp-content/plugins/progress-planner/classes/class-base.php:88
[...]

PR #400, opened 2025-04-02
Adjust timing for creating task provider objects

ilicfilip:
## Context

This PR is an alternative approach for the problem we have described in https://github.com/ProgressPlanner/progress-planner/pull/398 (both translation & adding task providers from 3rd party plugins).

In both cases the root cause is the more or less the same:
1. translatable strings needs to be added after the `init` hook
2. we have to wait for `plugins_loaded" in order to add 3rd party task providers

@aristath and I discussed it on Slack and it was decided not to define translatable strings in the task provider constructor, but to create get methods for them. I also took this opportunity to do the same for `Repetitive` tasks (since we had them still in get_details method) and move all shared methods and properties to the base `Local_Tasks` class.

For 2. I decided to delay adding 3rd party task providers, in the `Local_Tasks_Manager` class, to the `plugins_loaded` hook - as there were still problems when entire plugin init was done on that hook and this approach lets us use the rest of the plugin earlier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+= 1.1.2 =
+
+Under the hood:
+* Changed how the titles and descriptions of one_time and repetitive tasks are defined to be compatible with WP 6.8.
+
 = 1.1.1 =
 
 Bugs we fixed:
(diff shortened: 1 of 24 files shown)


aristath (changes requested):
Needs a changelog entry, and an accompanying PR in the https://github.com/ProgressPlanner/developer repo 👍

aristath (approved):

PR #405, opened 2025-04-03
POC - User can create "Remind me to review content" task from post edit screen

ilicfilip:
This is just a POC of what can be done now when we have refactored Todos into User tasks.

This PR adds a button on the Post edit screen which generates a new User (todo) task as reminder to review post content in 30 days. The task will not show up in the widget until 30 days pass.

For the purpose of this video I have changed the `available_at` param to be `time()` (without the extra 30 days), so task is visible immediately.

https://github.com/user-attachments/assets/0e545c85-229c-43f9-871f-307e748d66a9


aristath:
If we do that, we should make the TODOs snoozable - not for users but in the backend 👍 
Then, we should probably hide the task until the time comes for it to be unsnoozed

aristath:
Reminders should not be hardcoded to 30 days. User should be able to choose if they want a month, 3 months, 6 months, or a year.
When they _change_ the reminder, we should also be checking if there's already another reminder for that post, and if so then update the existing reminder instead of creating a new one.

ilicfilip:
Closing this one as not relevant anymore.

PR #406, opened 2025-04-03
Move notification counter to the bottom

ilicfilip:
## Context
Recently we have added the notification counter in the Progress Planner WP menu item, the counter was aligned to the top and it partly covered the Ravi's head and menu text:

Screenshot 2025-04-03 at 13 20 52

This PR moves the counter to the bottom, making the icon more visible:

Screenshot 2025-04-03 at 13 18 04


aristath:
I like it a lot more in the bottom... cc @Mijke

PR #419, opened 2025-04-09
Autoload data collectors

ilicfilip:
## Context

We use data collectors for some time now, they are primarily used in Local Tasks but can also be used by other parts of the codebase. They have their own hooks and data collector manager also takes care that their data is re-cached once a day.

This PR autoloads all data collectors, so we don't need to do it manually.

@aristath , I looked int making an object instance at the data collector class file (and then just required it) but I need the object in order to loop all of them and do daily cache refresh - so I created objects in the data manager class.


ilicfilip:
Closing this one as not relevant anymore.

PR #423, opened 2025-04-11
Change Create post task

ilicfilip:
## Context

This PR changes `Create post` task to following:

- Task is added to pending tasks only if there are no long posts published in the current week
- Task is not displayed in the RR widget when it's status is pending
- If there is pending `Create post`task and user creates long post, task will be added to the RR widget in order to be celebrated


ilicfilip:
Closing this one in favor of new specs: https://github.com/ProgressPlanner/progress-planner-pro/issues/111

PR #424, opened 2025-04-14
Add golden task to the weekly email

ilicfilip:
## Context

Add info about golden (todo) task to the get-stats REST API endpoint.

Necessary for https://github.com/ProgressPlanner/progress-planner-pro/issues/102


aristath (approved):
I don't particularly like very short var-names like `$r`, but in this case it's clear what it is and what it does so we can let it slide... 

LGTM 👍

PR #430, opened 2025-04-15
Add setting to determine which post types are considered 'valuable content'

ilicfilip:
Implements https://github.com/ProgressPlanner/progress-planner-pro/issues/112


classes/admin/class-page-settings.php
```diff
@@ -191,6 +192,20 @@ public function save_settings() {
...
+	 *
+	 * @return void
+	 */
+	public function save_post_types() {
+
+		$include_post_types = isset( $_POST['prpl-post-types-include'] ) // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			? array_map( 'sanitize_text_field', \wp_unslash( $_POST['prpl-post-types-include'] ) ) // phpcs:ignore WordPress.Security.NonceVerification.Missing, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			: array_intersect( [ 'post', 'page' ], \progress_planner()->get_settings()->get_public_post_types() ); // If no post types are selected, use the default post types (post and page can be deregistered).
+
+		\progress_planner()->get_settings()->set( 'include_post_types', $include_post_types );
```
aristath, line 206: I think it would be OK here to use `check_admin_referer`...
I know we already use it in the `store_settings_form_options` method, but if we want this method to be public then we should ensure it can't be abused.

Alternatively, we could just make this method `private` 🤔 

```suggestion
		// Check the nonce.
		\check_admin_referer( 'progress_planner' );

		$include_post_types = isset( $_POST['prpl-post-types-include'] )
			? array_map( 'sanitize_text_field', \wp_unslash( $_POST['prpl-post-types-include'] ) ) // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
			// If no post types are selected, use the default post types (post and page can be deregistered).
			: array_intersect( [ 'post', 'page' ], \progress_planner()->get_settings()->get_public_post_types() );

		\progress_planner()->get_settings()->set( 'include_post_types', $include_post_types );
```

aristath (changes requested):

ilicfilip:
@aristath , I have tested it and merged it

Should we add the same for `save_license` and `save_settings` methods?

aristath:
> Should we add the same for save_license and save_settings methods?

Yes please 👍

ilicfilip:
Done.

aristath (approved):

PR #447, opened 2025-04-28
Remove code for remote tasks

aristath:
## Context

So far we haven't used remote-tasks, and the separation we currently have for local/remote tasks has overcomplicated the implementation.

This PR removes all code for remote tasks, simplifying things.
When we do need them, we should implement them anew, taking into account the new structural changes that have happened in the past few releases.


ilicfilip:
I don't get any errors, but a lot of thing has changed so I will have to test in more detail.

One thing I found is that the users who upgrade from < v1.11 will get all tasks wiped out.

v1.1.1 migration script should migrate their data from old format (where we had pending tasks in separate wp_option entry) to `local_tasks`, but with changes in this PR they are migrated directly to [tasks](https://github.com/ProgressPlanner/progress-planner/blob/06f1be3cbf9ceb650df8acff683104f0156ef9d2/classes/update/class-update-111.php#L52-L66)

Then later in the upgrade process, when v1.4.0. does it's thing it [will populate](https://github.com/ProgressPlanner/progress-planner/blob/06f1be3cbf9ceb650df8acff683104f0156ef9d2/classes/update/class-update-140.php#L35) `tasks` entry with `local_tasks`, which for these users doesnt exist (thus it will be just be empty array and all task data will be wiped out).

I have tested this scenario and it is confirmed.

aristath:
@ilicfilip good catch! Should be fixed now 👍

ilicfilip:
I have checked PR in more detail and pushed 2 commits:

- First one fixes fatal error when `new Progress_Planner\Suggested_Tasks\Remote_Tasks\Remote_Task();` - mapping in the [autoload.php](https://github.com/Emilia-Capital/progress-planner/blob/e04f80ce8e2f58fef3f543dd5a3cb6fa192490f1/autoload.php#L88) is correct but `Tasks` class has required param which `Remote_Task` didnt have (small issue, since there is ~ 0% chance that anybody used our Remote tasks implementation)
- The second one is a bit bigger, `md5()` expects string (not array) so it actually causes PHP Fatal error during the v1.4.0 migration

Looking at the code, it looks like that updating for users < v1.1.1 will work fine.

But I think, and might be wrong, that now we have the same issue we had yesterday just for the users who are updating from v1.2.0. For them [this will be empty array](https://github.com/Emilia-Capital/progress-planner/blob/e04f80ce8e2f58fef3f543dd5a3cb6fa192490f1/classes/update/class-update-130.php#L74) (they still have all tasks in `local_tasks` entry, since for them v1.1.1 migration has ran before this update).

I am not sure if I am missing something, but can't we use `local_tasks` option entry [here](https://github.com/Emilia-Capital/progress-planner/blob/e04f80ce8e2f58fef3f543dd5a3cb6fa192490f1/classes/update/class-update-111.php#L52) and [here](https://github.com/Emilia-Capital/progress-planner/blob/e04f80ce8e2f58fef3f543dd5a3cb6fa192490f1/classes/update/class-update-111.php#L66) in the v1.1.1 migration script. Same for the v1.3.0 - [here](https://github.com/Emilia-Capital/progress-planner/blob/e04f80ce8e2f58fef3f543dd5a3cb6fa192490f1/classes/update/class-update-130.php#L74) and [here](https://github.com/Emilia-Capital/progress-planner/blob/e04f80ce8e2f58fef3f543dd5a3cb6fa192490f1/classes/update/class-update-130.php#L123). That way we have consistency between versions, as option name hasn't changed. And then rename the option in migration for v1.4.0?

PR #451, opened 2025-04-29
Fix repetitive dismissable tasks

ilicfilip:
## Context

This one is interesting, at least that it wasn't noticed so far.

Currently the only repetitive dissmisable task that we have is `Review content`, we check post_modified time to see if the posts has been updated in the previous 6 months - and if not, we create a task for the user.

There is also a possibility that user marks task as dismissed, in case the post is checked but it doesnt need to be updated, which also earns user a point and marks task as the completed.

Task data for this type of tasks stores the date (year & week) when the task was created and this is used not to suggest the same task again to the user in the current week.

So what happens if the user dismisses the task (marks it as completed) without editing the post?
The task will be again created in the following week, suggesting user to update the same post - which is a bug.

Similar case is when task is snoozed, but we have a check in place for that. The check for that one is looping through all the tasks, checking which belong to this task provider and their status and etc. It's not that efficient, so this is the reason why a new implementation was to cover dismissing tasks.

Also, for the upcoming v1.4 we have 2 more repetitive and dismissable tasks - Do Cornerstone and Do Orphaned content workout, which should be done by the user every 6 months. This implementation covers those as well.

From the currently opened PRs, this one is affected https://github.com/ProgressPlanner/progress-planner/pull/426 as it changes update frequency for the Review content task.

### DIFF
--- a/classes/class-suggested-tasks.php
+++ b/classes/class-suggested-tasks.php
@@ -633,9 +633,6 @@ class Suggested_Tasks {
 				// Insert an activity.
 				$this->insert_activity( $task_id );
 				$updated = true;
-
-				// Allow other classes to react to the completion of a suggested task.
-				\do_action( 'progress_planner_task_dismissed', $task_id );
 				break;
 
 			case 'pending':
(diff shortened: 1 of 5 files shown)


classes/class-suggested-tasks.php
```diff
@@ -633,6 +633,9 @@ public function suggested_task_action() {
 				// Insert an activity.
 				$this->insert_activity( $task_id );
 				$updated = true;
+
+				// Allow other classes to react to the completion of a suggested task.
+				\do_action( 'progress_planner_task_dismissed', $task_id );
```
aristath, line 638: Instead of doing this, it would probably be better if we do it _after_ the `switch`, and use an action-name like `"progress_planner_task_action_$action"`.
This way it will be possible to allow hooking in more action if and when it's necessary.

aristath (changes requested):

aristath (approved):

PR #452, opened 2025-04-29
Simplify the abstraction for one-time & repetitive tasks

aristath:
## Context

This PR simplifies the way we define new task-providers, and increases consistency.
We no longer need to extend different classes and use different implementations depending on whether the task is an one-time, or a repetitive task.

* Removes the `One_Time` & `Repetitive` classes
* Adds an `is_repetitive` property to task-providers
* Migrates logic from the old `One_Time` & `Repetitive` classes to the parent `Tasks` class, with conditions depending on the value of the new `is_repetitive` property
* Simplifies folder structure for providers, removing the one-time/repetitive distinction


aristath:
NOTE: review the documentation repo and update the code there once this gets merged

ilicfilip:
In general it looks good, but it needs to be checked in more detail - I wasn't getting any errors but "Review Content" tasks weren't marked as completed when post was updated because of the [this check](https://github.com/ProgressPlanner/progress-planner/pull/452/commits/522c97feb0e1b37da44391e0b752500a1ad3c520).

I will probably check it more later in the evening.

ilicfilip (approved):

Issue #433, opened 2025-04-15
Consolidate get_post_types_names() helper


ilicfilip:
Currently we are changing a lot of things related to content tracking & tasks.

In conversation with @aristath we mentioned that [Content_Helpers](https://github.com/ProgressPlanner/progress-planner/blob/ec9bca87698da673fafaeec68a3311a0e1d235f9/classes/activities/class-content-helpers.php) class will probably be remove as not needed anymore.

In parallel we are adding `valuable content` settings and tasks in following PRs:
- https://github.com/ProgressPlanner/progress-planner/pull/428
- https://github.com/ProgressPlanner/progress-planner/pull/430
- https://github.com/ProgressPlanner/progress-planner/pull/432

For now I have copied Content_Helpers::get_post_names() to [class-settings.php](https://github.com/ProgressPlanner/progress-planner/blob/86fa64de2d34f4939cad429d8494b37950059eaa/classes/class-settings.php#L120) in one of the branches so that helper is not removed if Content_Helpers is deleted, but we should find a better place for it once refactor is done (and also use it in all above PRs consistently).

Using just `\progress_planner()->get_settings()->get( [ 'include_post_types' ], [ 'post', 'page'] )` gets the saved value properly, but since CPT properties (like `public`) can be changed dynamically I think it's better to have a helper in place which filters out CPTs which are no longer eligible.

ilicfilip:
Done in #776, merged to `develop` as 9466fe45.

`Settings::get_post_types_names()` is now the single implementation. The six remaining `Content_Helpers` call sites point at it, and `Content_Helpers::get_post_types_names()` is a deprecated shim so external callers keep working.

That answers the "find a better place for it once the refactor is done" part of this issue: `Settings` is the home, and it keeps the CPT filtering the issue asked for — it intersects saved post types against currently-public ones, so post types that are no longer eligible get filtered out rather than being returned from stale settings.

The original #467 was closed in favour of #776; it had drifted ~2300 commits behind `develop` and consolidated in the opposite direction, into `Content_Helpers`.

PR #460, opened 2025-05-05
No need for word counts anymore

aristath:
## Context
This PR removes the word-counts code.
When calculating points for a post, we no longer take into account the post-length.


ilicfilip:
Looks good 👍 

@aristath , can we do this one now as well (if the refactor of Content Helpers class is complete): https://github.com/ProgressPlanner/progress-planner/issues/433 ?

aristath:
@ilicfilip yep, we should definitely consolidate those and ideally remove the Content_Helpers class if it's no longer necessary 👍

PR #467, opened 2025-05-05
Consolidate get_post_types_names() helper

ilicfilip:
Implements https://github.com/ProgressPlanner/progress-planner/issues/433


ilicfilip:
Closing in favour of #776, which implements the same consolidation against current `develop`.

This branch had drifted ~2300 commits behind `develop` and conflicted in 6 of its 7 files. More importantly, it consolidates in the direction the codebase has since moved away from: it deletes `Settings::get_post_types_names()` and moves the implementation into `Content_Helpers`, but `develop` has since settled on `Settings` as the canonical home — that's where the newer suggested-task and data-collector classes call, and where the exclusion list has been maintained. Rebasing would have meant reverting that decision.

The underlying issue is still real, so #433 stays open until #776 merges. #776 goes the other way: it points the remaining `Content_Helpers` call sites at `Settings` and leaves `Content_Helpers::get_post_types_names()` as a deprecated shim.

PR #481, opened 2025-05-09
Fix snoozing of some Yoast SEO tasks

ilicfilip:
This PR fixes snoozing of Yoast SEO recommendations that we added in v1.4.0.

Unrelated bug, which doesnt affect users for now, is that Workout tasks were missing `$is_repetitive = true;`, most likely introduced when Task structure was changed (and we have removed `One_Time` and `Repetitive` classes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+= 1.4.2 =
+
+Bugs we fixed:
+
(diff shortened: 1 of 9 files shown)


aristath (approved):
Code LGTM.
The logic makes sense 👍

PR #485, opened 2025-05-12
Don't redirect user to Progress Planner dashboard if 'redirect_to' GET or POST parameter is set

ilicfilip:


classes/class-base.php
```diff
@@ -465,6 +465,10 @@ public function redirect_on_login( $user_login, $user ) {
 			return;
 		}
 
+		if ( ( isset( $_POST['redirect_to'] ) && '' !== $_POST['redirect_to'] ) || ( isset( $_GET['redirect_to'] ) && '' !== $_GET['redirect_to'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.NonceVerification.Missing -- We're not processing any data.
```
jdevalk, line 468:
```suggestion
		if ( isset( $_REQUEST['redirect_to'] ) && '' !== $_REQUEST['redirect_to'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.NonceVerification.Missing -- We're not processing any data.
```

ilicfilip: I intentionally used both GET and POST to be explicit about it, but sure - makes sense

aristath (approved):

PR #490, opened 2025-05-23
CPT JS refactor

ilicfilip:
Since we switched to CTP approach we leverage WP REST API a lot, which means that we need to wait for API client to ready  before doing almost anything. Because of that we have added `wp.api.loadPromise.done` checks in a lot of places (and we will add more).

For example, sometimes on page load Todo widget gets stuck with `Loading items...` message (although I can add new tasks), the on on page reaload it doesn't happen.

This is not reliable as basically as very often something is not working consistently and we're constantly fighting racing conditions (ie adding more `wp.api.loadPromise.done` checks or setTimeouts).

This PR tries to solve it by initing the Tasks widget code (RR and todo) when API client is ready, so `wp.api.loadPromise.done` is checked only once. It loads sets global `window.progressPlannerSuggestedTasksTerms` before other code is executed, so we dont need to check constantly if it set or not.

The other issue is that most the functionality is handled through events, but this led us to nested (chained) events where event.details could have a param which tells the event callback to trigger another event, which then can trigger another.. which is very difficult to follow through and debug.

To solve that this PR adds asyncEvent implementation, so instead of passing callbacks we can do:

```
// Inject published tasks from this category.
prplDispatchAsyncEvent( 'prpl/suggestedTask/injectCategoryItems', {
	category,
	status: 'publish',
} )
	.then( ( data ) => {
		data.forEach( ( item ) => {
			document.dispatchEvent(
				new CustomEvent( 'prpl/suggestedTask/injectItem', {
					detail: item,
				} )
			);
		} );
	} )
	.then( () => {
		// Update the UI.
		prplSuggestedTasksToggleUIitems();
	} );
```

Triggering event in the usual way still works, ie: 

```
document.dispatchEvent(
	new CustomEvent(
		'prpl/suggestedTask/injectCategoryItems'',
		{
			detail: ...,
		}
	)
);
```

But hopefully using promises will be more readable and add more flexibility (which we need, since 2 widgets work similar but not the same).

This PR is still a WIP, so here are some other bugs that it is trying to fix:

- [ ] [Global var for user tasks](https://github.com/ProgressPlanner/progress-planner/blob/07b952daad2d965cb472312a7f2cf5f181c8346f/assets/js/widgets/todo.js#L15-L17) is not properly populated, quick fix (on page load) was added in [this commit](https://github.com/ProgressPlanner/progress-planner/commit/07b952daad2d965cb472312a7f2cf5f181c8346f), but it is not updated down the line (when user task is added, removed, edited).
- [x] [Here](https://github.com/Emilia-Capital/progress-planner/blob/c1820ed4eb01ae6afc3a412069a75b584c643973/assets/js/web-components/prpl-suggested-task.js#L539-L542), when the task's checkbox is checked, we set status to `pending_celebration` even to user tasks.
- [ ] Re-injecting tasks from the same category after it is marked as completed doesn't work

### DIFF
--- a/assets/js/celebrate.js
+++ b/assets/js/celebrate.js
@@ -8,9 +8,9 @@
  */
 /* eslint-disable camelcase */
 
-// Create a new custom event to trigger the celebration.
-document.addEventListener( 'prpl/celebrateTasks', ( event ) => {
-	wp.api.loadPromise.done( () => {
+window.prplInitCelebrate = () => {
+	// Create a new custom event to trigger the celebration.
+	document.addEventListener( 'prpl/celebrateTasks', ( event ) => {
 		const postsCollection = new wp.api.collections.Prpl_recommendations();
 		postsCollection
 			.fetch( {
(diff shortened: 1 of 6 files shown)


aristath:
@ilicfilip Why do we _need_ to wrap our code in `init` functions? 🤔 
I don't see a reason

PR #496, opened 2025-05-30
Send email (interactive) task tweaks

ilicfilip:
For now:

- prefix ajax_action 
- better catch various sending failures

To test add this to functions.php:

```
add_action( 'phpmailer_init', function( $phpmailer ) {
    // $phpmailer->to = ''; // Triggers a PHP error.
    // $phpmailer->clearAddresses(); // Clears all "to" addresses, sending fails but without PHP error.
    // $phpmailer->Mailer = 'smtp'; // Simulate misconfigured SMPT server, sending fails without PHP error.
} );
```

As is now, if there is an error user will get to the "Troubleshooting" step (same step as if he said that he didn't receive test email), but we can add different (more descriptive) step just for this case.

### DIFF
--- a/assets/css/page-widgets/suggested-tasks.css
+++ b/assets/css/page-widgets/suggested-tasks.css
@@ -395,6 +395,16 @@
 					height: 100%;
 				}
 			}
+
+			&.prpl-note-error {
+				color: #9f0712;
+				background-color: var(--prpl-background-red);
+
+				.prpl-note-icon {
+					color: var(--prpl-color-notification-red);
+				}
+
+			}
 		}
 
 		/* To align the buttons to the bottom of the column. */
--- a/assets/css/page-widgets/suggested-tasks.css
+++ b/assets/css/page-widgets/suggested-tasks.css
@@ -147,6 +147,13 @@
 		}
 	}
 
+	.prpl-suggested-task {
+
+		.prpl-suggested-task-checkbox {
+			flex-shrink: 0; /* Prevent shrinking on mobile */
+		}
+	}
+
 	/* Disabled checkbox styles. */
 	.prpl-suggested-task-disabled-checkbox-tooltip {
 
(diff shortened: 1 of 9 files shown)


Mijke:
@aristath @ilicfilip 
@tacoverdo asked @irisgue and me to have a look at the copy and which text should be on the left and which should be on the right. 

We ended up with this: 
https://www.figma.com/proto/YGmG3KOj7jfQz5KUbUAbk3/interactive-tasks?node-id=4444-2137&t=WUp5s1m4NPXMy9cH-0&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=4%3A572&show-proto-sidebar=1

The copy has been tweaked a bit. I'll add it in the comment below

Mijke:
@ilicfilip @aristath 

Tweaked copy. @irisgue can you confirm this is all good?

### STEP 1 

**left**
**Test if your site can send emails**
Your WordPress site sometimes needs to send emails. For example, to reset a password, send a comment notification, or warn you when something breaks. Contact forms also use email.

It’s important to check if these emails are actually sent. Enter your email address on the right to get a test email.

**right**
Where should we send the test email?
Alert: You should get the email in a few minutes. In rare cases, it might take a few hours.
CTA: Next step >

### STEP 2 - failed

**left**
**We tried to send a test email**
We just tried to send the email “ [TITLE] to [Email address], but unfortunately it didn’t work.

**right**
Alert: The test email didn’t work. The error message was: [error message].

There are a few common reasons why your email might not be sending. Check the troubleshooting guide to find out what’s causing the issue and how to fix it.

CTA 1: < Try again
CTA 2: Retry later >

### STEP 2 - successful 

**left**
 **We sent a test email**
 We just sent the email [TITLE] to [Email address].
 
 **right**
 Did you get the test email?
Alert: In rare cases, it might take a few hours.

Radio buttons: 
Yes
No

CTA: Next step >

### STEP 3 - received

**left**
**Your email is set up properly!**
Great, you received the test email! This indicates email is set up properly on your website.

**right**
Celebrate this achievement!

CTA: Collect your point!

ilicfilip:
@tacoverdo , @irisgue , @mijke - all changes have been applied

PR #508, opened 2025-06-16
v1.6.0

aristath:


tacoverdo:
# Trigger-happy confetti
Confetti fires before "installation/upgrade" modal has been closed. This might cause users to 'miss' their first celebration, as their focus isn't on the background.

https://github.com/user-attachments/assets/1a4e4906-c93d-4506-a7c0-1d137809b391

ilicfilip:
@tacoverdo , I will re-check upgrade but after installation we have onboarding process (which doesn't include modal) and confetti fire properly - modal you're seeing is specific to Playground.

tacoverdo:
# Magic task completion when setting a page in the PP settings
A review task gets magically completed without touching the page when it's set as a page in the settings. 

This is counterintuitive to my expectation that a review task would only be completed when actually editing the post/page. 

https://github.com/user-attachments/assets/b80c2c2a-f775-4d86-9d4f-d96e1818d6d9

PR #511, opened 2025-06-16
API to allow branding changes

aristath:


classes/ui/class-branding.php
```diff
@@ -0,0 +1,123 @@
...
+		/**
+		 * Filter the strings used for branding purposes.
+		 *
+		 * @param array $strings The strings.
+		 * @return array
+		 */
+		return \apply_filters(
+			'progress_planner_branding_strings',
+			[
+				'ravi' => __( 'Ravi', 'progress-planner' ),
```
ilicfilip, line 119: @aristath , what do you think about including `'s`  in the string as well - it is used in 3 places and all of them use `'s`, also without it strings are most likely not translatable properly?

Ie, in Serbian Рави's doesnt make sense (probably in Greek and Dutch as well)

ilicfilip:
`Page::enqueue_styles` was called twice on WP Dashboard page, from [here](https://github.com/Emilia-Capital/progress-planner/blob/07a019eba8ed72042bf037cbc074440dbdae26b6/classes/admin/class-dashboard-widget-score.php#L40) and [here](https://github.com/Emilia-Capital/progress-planner/blob/a537009bd2a85805acf4a2508e7303413aae7d87/classes/admin/class-dashboard-widget-todo.php#L39), so I have added static var to prevent that from happening.

PR #513, opened 2025-06-17
Allow excess points to complete previous monthly badge

aristath:
## Context
This will help badge-competitive people to work on their site and not be demotivated in case they miss a badge by a few points.

Text: 
> Congratulations! You have completed more tasks than the target score. The additional (X) points will help you complete the previous monthly badge that you missed.

Screenshot 2025-06-17 at 13 50 21

### DIFF
--- a/classes/admin/widgets/class-suggested-tasks.php
+++ b/classes/admin/widgets/class-suggested-tasks.php
@@ -75,6 +75,8 @@ final class Suggested_Tasks extends Widget {
 	 * @return void
 	 */
 	public function enqueue_scripts() {
+		parent::enqueue_scripts();
+
 		// Enqueue the badge scroller script.
 		\progress_planner()->get_admin__enqueue()->enqueue_script(
 			'widgets/suggested-tasks-badge-scroller',
--- a/classes/admin/widgets/class-suggested-tasks.php
+++ b/classes/admin/widgets/class-suggested-tasks.php
@@ -7,6 +7,7 @@
 
 namespace Progress_Planner\Admin\Widgets;
 
+use DateTime;
 use Progress_Planner\Badges\Monthly;
 
 /**
@@ -24,7 +25,7 @@ final class Suggested_Tasks extends Widget {
 	/**
 	 * Get the score.
 	 *
-	 * @return int The score.
+	 * @return array<string, int> The scores.
 	 */
 	public function get_score() {
 		$activities = \progress_planner()->get_activities__query()->query_activities(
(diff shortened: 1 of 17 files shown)


classes/admin/widgets/class-suggested-tasks.php
```diff
@@ -40,7 +41,22 @@ public function get_score() {
...
+	}
+
+	/**
+	 * Get previous month badge.
+	 *
+	 * @return \Progress_Planner\Badges\Monthly|null
+	 */
+	public function get_previous_month_badge() {
+		return Monthly::get_instance_from_id(
+			Monthly::get_badge_id_from_date( ( new DateTime() )->modify( 'first day of next month' ) )
```
ilicfilip, line 58: shouldn't this be "first day of the previous month"?

if so, we need a check if that badge exists (for example, I activated plugin this month)

views/page-widgets/suggested-tasks.php
```diff
@@ -46,21 +46,38 @@
...
 		<span id="prpl-widget-content-ravi-points-number" class="prpl-widget-content-points-number">
-			<?php echo (int) $prpl_widget->get_score(); ?>pt
+			<?php echo (int) $prpl_widget->get_score()['target_score']; ?>pt
 		</span>
 	</div>
 
+	<?php
+	if (
+		$prpl_widget->get_score()['score'] > $prpl_widget->get_score()['target_score']
+		&& $prpl_widget->get_previous_month_badge()->progress_callback()['progress'] < 100
```
ilicfilip, line 67: Depends on how previous month badge is implemented, we also might need a check here to see if the badge for the previous month exists.

ilicfilip (changes requested):

classes/admin/widgets/class-suggested-tasks.php
aristath: > shouldn't this be "first day of the previous month"?

Good catch. Fixed.

> if so, we need a check if that badge exists (for example, I activated plugin this month)

Already checked in the `get_instance_from_id()` method, returns null if the badge doesn't exist

aristath:
@ilicfilip A lot of things changed in this PR since you last reviewed it, could you please do another pass at it? 🙏

ilicfilip:
@aristath , monthly badge on WP Dashboard screen is for some reason stuck at 1:
Screenshot 2025-06-20 at 10 51 18

ilicfilip (approved):

PR #520, opened 2025-06-23
Split the suggested-tasks widget and double its width

aristath:
Fixes https://github.com/ProgressPlanner/progress-planner-pro/issues/123, https://github.com/ProgressPlanner/progress-planner-pro/issues/125, https://github.com/ProgressPlanner/progress-planner-pro/issues/156


aristath:
Fixed. The issue was that the grid was not getting re-calculated when celebrated items were being removed.

aristath:
Implemented the action links in https://github.com/ProgressPlanner/progress-planner/pull/520/commits/97631c6e309cdcadc1daa1f23911288ea9f7ebab
However, we'll have to go over all tasks and figure out the text that will actually be displayed for each one (see https://github.com/ProgressPlanner/progress-planner-pro/issues/156)

cc @tacoverdo

aristath:
Note to self: Using `git-bisect` it looks like https://github.com/ProgressPlanner/progress-planner/pull/520/commits/16188492c447c85829de8041d8bf525a1bf3c5e3 broke the popovers layout, need to take a look at that

ilicfilip:
@aristath , looks like a different commit was the cause - the problem was that the interactive popover styles were moved inside the `..prpl-suggested-tasks-list` but popover html is not injected there.

Now there is another issue, `.radio` and ` .prpl-column (on hover)` have hover applied from [here](https://github.com/Emilia-Capital/progress-planner/blob/a4deb39abfffcb829f68932108044eaa5e128034/assets/css/page-widgets/suggested-tasks.css#L258-L260). I can't untagle what happened there, if that was a merge from the dev branch or you added them for a reason, but on dev branch those styles should be applied to [the button only](https://github.com/ProgressPlanner/progress-planner/blob/090936975bd10efadae386a18d8125bb9c4c1383/assets/css/page-widgets/suggested-tasks.css#L459C4-L471).

Mijke:
@aristath @ilicfilip @tacoverdo not sure if this is on the radar, but for visibility a link to the latest changes in the to-do list
https://github.com/ProgressPlanner/progress-planner-pro/issues/125#issuecomment-3191235017

aristath:
Yeah, the order of actions has not been fixed yet. We refactored and abstracted the way we add them, so each action now has a priority we can tweak. That is the next step in this PR for Monday 👍

PR #530, opened 2025-06-26
v1.7.0

aristath:


tacoverdo:
# Check site locale
This recommendation now links to the admin page to set the site language. And because we think it needs changing, it has the +1 icon in the WP admin settings. However, in this case the website is written fully in Dutch, so the site setting 'Dutch' is indeed correct, despite my browser reporting a en_US preference.

By turning this task into an interactive one, we can ask the user about the (primary) language on the site, and also reward them for _not_ changing the setting (when it's already correct).

PR #538, opened 2025-07-01
New task provider: Disable comment pagination

aristath:
Fixes https://github.com/ProgressPlanner/progress-planner-pro/issues/136

This provider required adding a new dependencies system, so we can show/hide tasks depending on the status of other tasks.

NOTE: To test this on Playground, you'll need to:
* Go to Settings -> Discussion
* Enable comment pagination
* PRPL Debug -> Delete -> Delete Suggested Tasks

### DIFF
--- a/assets/js/recommendations/disable-comment-pagination.js
+++ b/assets/js/recommendations/disable-comment-pagination.js
@@ -11,5 +11,5 @@ prplInteractiveTaskFormListener.settings( {
 	settingPath: '{}',
 	taskId: 'disable-comment-pagination',
 	popoverId: 'prpl-popover-disable-comment-pagination',
-	settingCallbackValue: () => false,
+	settingCallbackValue: () => '',
 } );


tacoverdo:
I now see the task in the task list. However, it sends me to the settings page to uncheck a box. Shouldn't this be an interactive task to make that process easier?

ilicfilip:
@tacoverdo , I will convert it to the interactive task and we can merge it before the release

ilicfilip:
@aristath I have converted task to be interactive, but I had to register setting to be available in the REST API since that isn't [done by WP core](https://github.com/WordPress/WordPress/blob/master/wp-includes/option.php#L2715-L2943) (like the options we used with interactive tasks so far).

We just need to finalize the task description (since currently it is the same as the task title) and popover content
Screenshot 2025-08-08 at 16 03 48

aristath:
> @aristath I have converted task to be interactive, but I had to register setting to be available in the REST API since that isn't [done by WP core](https://github.com/WordPress/WordPress/blob/master/wp-includes/option.php#L2715-L2943)

I reverted that change and implemented it using `prplInteractiveTaskFormListener.settings` instead of `prplInteractiveTaskFormListener.siteSettings` 👍 

Tested it and it's still working as expected, could use a 2nd check to be sure

ilicfilip (approved):
@aristath , I have just changed the option value to be an empty string ('false' was saved in as the option value in the DB)

PR #540, opened 2025-07-03
Add new task provider: Select site locale

aristath:
Checks if the browser language is not the same as the site locale, and adds a new task in case they differ.

Fixes https://github.com/ProgressPlanner/progress-planner-pro/issues/135


classes/suggested-tasks/providers/class-select-locale.php
```diff
@@ -0,0 +1,144 @@
...
+	 */
+	protected function get_locales() {
+		$cached = \progress_planner()->get_utils__cache()->get( 'all_locales' );
+		if ( $cached ) {
+			return $cached;
+		}
+
+		$response = \wp_remote_get( 'https://api.wordpress.org/translations/core/1.0/' );
+		if ( \is_wp_error( $response ) ) {
+			return [];
```
ilicfilip, line 120: cache empty array for 5 mins like we do in https://github.com/Emilia-Capital/progress-planner/blob/e0af996a7bc0ac478007dcbd6002ed6f46d91fc2/classes/class-lessons.php#L64-L67 ?

ilicfilip (approved):

fleurterpstra:
@aristath I tried to test this to see the flow of the task, but it's not working. How can I test this?

aristath:
The task only shows-up when the browser is not using the same language as the site.
So for example if you have your browser in Dutch, but your site is in English 😉

PR #556, opened 2025-07-22
Select site timezone task

ilicfilip:
Based on: https://github.com/ProgressPlanner/progress-planner-pro/issues/71

**Note:** Description needs to be adjusted probably.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 = 1.7.0 =
 
+Added these recommendations from Ravi:
+* Select site timezone.
+
 Bugs we fixed:
 
 * Fix issue where "Perform all updates" task was incorrectly marked as completed.
--- a/classes/suggested-tasks/class-task.php
+++ b/classes/suggested-tasks/class-task.php
@@ -90,11 +90,13 @@ class Task {
 	 * @return void
 	 */
 	public function delete(): void {
-		$this->data = [];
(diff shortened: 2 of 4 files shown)


aristath:
Code looks good 👍

tacoverdo:
Right now, the task is sending people to a generic settings page where they have to figure out where to find the time zone setting. As an interactive task, we can guide the user to setting the correct time zone for their site.

Without additional context, I'm not sure our users will know why it matters what time zone a site is set to. And I'm confident that by using an interactive task, we can take a more guided approach to this setting. 

So I'm wondering, wouldn't it make sense to make this an interactive task?

ilicfilip:
Hmm, something is off - link should also make the browser scroll down to the setting (like it is done for "set site icon"), not just open the Settings page.

As far I can see the Ravi icon is there, I am checking why the rest is not working.

ilicfilip:
As far I can see we can use this snippet to output the `<select>` element for the Select timezone in the interactive task popup: https://github.com/WordPress/WordPress/blob/master/wp-admin/options-general.php#L360-L389

But first we should merge https://github.com/ProgressPlanner/progress-planner/pull/544, since it already has an implementation for converting other tasks to interactive.

ilicfilip:
This task is also now interactive, but we need to finalize task description and the popover content
Screenshot 2025-08-08 at 16 36 27

aristath (approved):

PR #582, opened 2025-08-21
Split streaks widget

aristath:
Fixes https://github.com/ProgressPlanner/progress-planner-pro/issues/128


ilicfilip (changes requested):
I see some minor spacing issues, like space between the badge and it's name (it's much smaller in the new design), but that is also adjusted in https://github.com/ProgressPlanner/progress-planner/pull/585 so it should be fixed once everything is merged in.

But there is something weird with the how widgets are positioned, even after the page load I can see them changing places couple of times:

https://github.com/user-attachments/assets/0e0b433d-9169-4c8a-afc0-9aa22426104c


Also looks like the same happens if I open open Dev tools, which lowers the viewport height but doesn't change it's width - so the widget order shouldn't be changed?

aristath:
> there is something weird with the how widgets are positioned, even after the page load I can see them changing places couple of times

Yeah, that's because I had to add a `setTimeout` in order to make things work properly in https://github.com/ProgressPlanner/progress-planner/pull/589

Can you think of a better solution for that? 🤔

ilicfilip:
Part of the problem is that we trigger many `prpl/grid/resize` events on page load (I counted 8 on my local, without celebration), from various places and some of which are also triggered using timeouts. I tried adding a debounce function, but as we need to trigger at least 2 events it doesnt fix the issue.

As far I can see, [this ](https://github.com/Emilia-Capital/progress-planner/blob/55ba6d22272f8cf0a5fd4cb0c0e4980661788b06/assets/js/grid-masonry.js#L130) the timeout in question, which to me it looks like is added to handle viewport resize?

ilicfilip (approved):

PR #585, opened 2025-08-22
Monthly badge and missed badge(s)

ilicfilip:


aristath:
> @aristath , should be fixed now!

Confirmed 👍

PR #586, opened 2025-08-26
Handle failed interactive task requests

ilicfilip:
When "complete" interactive task button is clicked in the background an AJAX request is fired, but until now we haven't checked if the request returned error or not. This might be confusing as even if the error occurred the task will be marked as completed and celebration triggered.

This PR adds a check for that and it will not mark task as completed, but we need to decide how exactly we want to display error to the user.

Also, it would be nice to add a spinner while request is in progress since some requests can take time (ie, set locale in the background downloads language packs). We have spinner added for the User (Todo) tasks, but we need to decide if want to use it in the same way (it covers task list and input field).


aristath (approved):

aristath:
A spinner can be added in a follow-up PR if we deem it necessary. For now this looks good so I'll go ahead and merge it 🥇

PR #587, opened 2025-08-26
New task provider: Check SPF, DKIM, DMARC

ilicfilip:
This PR is still in WIP state but it adds new task provider, to check email related DNS records and spam score - the goal is to let user check how likely is that his email will be delivered and what can be improved.

Postmark is used to check email DNS records and conveniently it also uses Spam Assassin to assign spam score to it.

It needs to be decided how exactly the task should work for the user, biggest problem is that email can be sent a bit later (not right after the user clicks on the "Check" button) - but for now it looks like this:

- User opens task popover and clicks on the check button
- In the background an AJAX request is fired and it's callback does following:
- pings our remote server to get the nonce
- pass the nonce and the unique email subject to the server (server should store subject for later check)
- sends an email and waits for 10s (postmark will trigger webhook after 10-15s)
- pings server again to check, passing the unique subject, to see if the postmark report is ready

Things that need to be implemented:

- Where to store unique subject, so we can pair it with the Postmark data once the webhook is triggered (ideally this is not `wp_options` table) - decided, **new SQLite database**
- Handle case when report is not ready after 10-15s. Ideally we display a message to the user to let him know not to close task popover yet and trigger new check every 10s.
- Handle the case when email is sent after an hour or so, user can't wait that long so we should find a way to let know user that check is done even after the task popover is closed.
- Think of the message which user gets when report is done, it's not unique (just pass or fail) since for example SPF can be set but DKIM not and etc.


ilicfilip:
Closing this one, we can revisit it in the future if needed.

Issue #564, opened 2025-08-11
[Recommendation] Set the date format


ilicfilip:
### Recommendation title

Set the date format

### Recommendation description

Nudge user to set the date format (Settings -> General), similar to what we did for the [set the timezone task](https://github.com/ProgressPlanner/progress-planner/pull/556)

### Recommendation category

configuration

### Onboarding recommendation

false

### Points to award

1

### Recommendation priority

low

### Dismissable

true

### Snoozable

true

PR #590, opened 2025-09-01
New task provider: Set the date format

ilicfilip:
Closes https://github.com/ProgressPlanner/progress-planner/issues/564


ilicfilip:
It turns out that the default date format also has it's translation in the language pack, so I have added that as a first choice (radio button) when setting the date format.

PR #591, opened 2025-09-01
Abstract implementation for actions in RR

aristath:
This PR abstracts the way we add actions to recommendations.
This will make it easier to add custom actions to items, as well as tweak their order.


aristath:
> We're injecting the actions using [add_data_to_rest_response](https://github.com/Emilia-Capital/progress-planner/blob/f6f0099a39e744eea48a4644fa29978318a41b40/classes/class-suggested-tasks.php#L558-L573) which is hooked into the rest_prepare_prpl_recommendations filter - but we already hook to the same filter and more or less do the same thing for the link [here](https://github.com/Emilia-Capital/progress-planner/blob/f6f0099a39e744eea48a4644fa29978318a41b40/classes/class-suggested-tasks.php#L448-L462). So maybe we can merge those functions (not to loop over all tasks and fetch provider for every one twice)?

Good point, I missed that. Pushed 👍

ilicfilip (approved):

PR #599, opened 2025-09-09
v1.8.0

aristath:


tacoverdo:
The main action of "Remove inactive plugins" links to the plugins page, making the user search for their inactive plugins. 

image

Linking directly to `/wp-admin/plugins.php?plugin_status=inactive` would only show the inactive plugins, which takes friction away from the user.

aristath:
> The 'add internal links' action item links to /wp-admin/admin.php?page=wpseo_tools#/fix-orphaned-content. It should link to /wp-admin/admin.php?page=wpseo_workouts#orphaned instead.

Is Yoast-SEO Premium installed on that site?
I'm wondering if perhaps the URL is only for premium... 🤔

ilicfilip:
> On every page load of our dashboard page, the list of recommendations looks broken before it 'corrects'. Is there a way for us to prevent this from happening?

PR -> https://github.com/ProgressPlanner/progress-planner/pull/614

aristath:
Fixed the issue with the points-rewarding in previous monthly badges

PR #605, opened 2025-09-11
Stop using post-meta to store a task's points

aristath:


ilicfilip:
There was a PHP notice when there were no User tasks:

```
Warning: Undefined array key "user" in htdocs/wp-content/plugins/progress-planner/classes/suggested-tasks/providers/class-user.php on line 118

Warning: foreach() argument must be of type array|object, null given in htdocs/wp-content/plugins/progress-planner/classes/suggested-tasks/providers/class-user.php on line 118
```

The commit I pushed fixes that.

Is update migration required, so task stays golden after the plugin update?

PR #619, opened 2025-09-16
Improve playground workflow

jdevalk:
Fix the zip file in the playground comment workflow to:

- Be called `progress-planner.zip` and expand to `progress-planner`, so it can override existing installs.
- Add ` - PR <pr number>` to the version number in the plugin.
- It removes the reliance on a php file on progressplanner.com and just uses the file `playground.json` in `.wordpress-org/blueprints/`. I've split it off specifically from the `blueprint.json` in that same directory so that we can define different things for our testing.


aristath (approved):

jdevalk:
CleanShot 2025-09-16 at 13 27 09

PR #622, opened 2025-09-17
Previous month points are not decreased when golden task is set back to pending

### LINKED ISSUE #621: Previous month points are not decreased when golden task is set back to pending
Earlier today we had an issue where the points for the previous month badge weren't updated when task was completed.

That is fixed, but they are not decreased once the golden task is set back from completed to pending:

https://github.com/user-attachments/assets/ab4c470a-a8d9-453b-b120-a7ca02b5bed0

ilicfilip:
Fixes #621

The problem was that the `span.number` element was removed after the first update, which then prevented [this code](https://github.com/Emilia-Capital/progress-planner/blob/f783dcab3ab1cd21dade374403638f47953b7d9a/assets/js/web-components/prpl-badge-progress-bar.js#L192-L195) to update it again

### DIFF
--- a/assets/js/web-components/prpl-badge-progress-bar.js
+++ b/assets/js/web-components/prpl-badge-progress-bar.js
@@ -113,10 +113,8 @@ const prplUpdatePreviousMonthBadgeProgressBar = ( pointsDiff ) => {
 	);
 
 	if ( remainingPointsEl ) {
-		remainingPointsEl.textContent = remainingPointsEl.textContent.replace(
-			remainingPointsEl.getAttribute( 'data-remaining' ),
-			badgeMaxPoints - badgeNewPoints
-		);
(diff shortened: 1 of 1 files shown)


ilicfilip:
@aristath , does [this code](https://github.com/Emilia-Capital/progress-planner/blob/f783dcab3ab1cd21dade374403638f47953b7d9a/assets/js/web-components/prpl-badge-progress-bar.js#L111-L124) should run only on the progress bar for the previous month (like it is now) or it should run for both months?
Or it should be removed now when we have separate function for updating?

The other issue I found is that when the Todo task is set back to pending to point is substracted from the gauge for the current month as well (of course it is corrected after the page reload)

https://github.com/user-attachments/assets/15c15ae2-f6d5-4b3e-a4a8-a78ebe5aa273

aristath (approved):

PR #625, opened 2025-09-17
Color palette refactor

aristath:
Fixes https://github.com/ProgressPlanner/progress-planner-pro/issues/157


aristath:
I see these (old) colours are still used:
* `--prpl-color-gray-3`
* `--prpl-color-gray-4`
* `--prpl-background-red`
* `--prpl-background-blue`

ilicfilip:
@aristath those are the colors for which we wait for feedback.

I left them on purpose because its easier for me to find (and replace) old color var which has just a few occurrences than new var which is used in a lot of places now.

Other occurrences for those color vars are already replaced.

Sorry for not pointing out that more in the message on Slack yesterday.

PR #626, opened 2025-09-17
Remove "task checker" from onboarding flow

ilicfilip:
This PR removes "Task checker" (check for completed tasks during onboard).


aristath:
Great!
Content-scanning was already removed in https://github.com/ProgressPlanner/progress-planner/pull/632 so I think we can move forward with this one as well 👍

ilicfilip:
@aristath , yes - I merged the changes and it looks like the tests finally passed!

(they worked locally always, but some timing was off on CI when the page reloads after onboarding)

PR #644, opened 2025-09-30
AIOSEO integration

jdevalk:
## Summary

* Adds a first few AIOSEO recommendations.

* Built with Claude, so please review well.

To test, install AIOSEO alongside this plugin.


ilicfilip:
The code looks good to me, I tried to break it but without success.

I have updated action links where possible, so after the AIOSEO page is opened user is scrolled to the section which has the option that has to be changed (I used the same URL parameters they use for scrolling and highlighting sections, so nothing custom is added).

`Remove post authors feeds` and` Disable global comment RSS feeds` are bit confusing to change (at least for me), since this is the page user gets:

Screenshot 2025-10-01 at 10 13 08

And the options are actually inside the "Crawl Cleanup" section - so maybe we can change them to interactive tasks?

I have also updated the Changelog, but that obviously needs to be checked once more when we create the external pages.

jdevalk:
Yeah an interactive task might be better, could you whip something up for that?

jdevalk:
Yeah i think that's better!

We might have to add a bit of text about why we suggest these things in the popup. Let me give you those:

For comment feeds (where we disable just the global one now but we should actually disable `postComments` too):

> We suggest disabling both the global "recent comments feed" from your site as well as the "comments feed" per post that WordPress generates. These feeds are rarely used by real users, but get crawled a lot. They don't have any interesting information for crawlers, so removing them leads to less bot-traffic on your site without downsides.

For author feeds:

> The author feed on your site will be similar to your main feed if you have only one author, so there's no reason to have it.

For media pages:

> WordPress creates a "page" for every image you upload. These don't add any value but do cause more crawling on your site, so we suggest removing those.

jdevalk:
And I forgot two:

Author archives:

> Your author archives are the same as your normal archives because you have only one author, so there's no reason for search engines to index these. That's why we suggest keeping them out of search results.

Date archives:

> Date archives rarely add any real value for users or search engines, so there's no reason for search engines to index these. That's why we suggest keeping them out of search results.

ilicfilip:
All done!

I have also update the ID (and everything relared) for the `Feed comments task`, since now it is for both global and post comments, from `aioseo-crawl-settings-feed-global-comments` to `aioseo-crawl-settings-feed-comments`

PR #648, opened 2025-10-02
Refactor gauges & badge progress bars

### LINKED ISSUE #623: Gauge / progress bar issues when golden task is set back to pending
Spin off from: https://github.com/ProgressPlanner/progress-planner/issues/621

- When the golden Todo task is set back to pending to point is substracted from the gauge for the current month as well, which shouldn't happen (it is correct after the page reload)
- The progress bar for the previous month (with the badge) doesn't move "back" when point is removed

https://github.com/user-attachments/assets/1652b054-096d-42cd-8f01-c38c7b4e81ef

ilicfilip:
Over the time gauges and badge progress bars became complex, far beyond what for what they initially were made for.
This is especially the case for the badge progress bars, which not only depend on the the monthly badge gauge but trigger multiple side effects (update elements which are not part of that component).

The goal of this PR is to introduce a simpler way to control those 2 components and make them more `reactive` (still using Vanilla JS).

That is done by following:

- Both gauge and badge progress bar components have state now, so DOM is not longer used to store data.
- Components still have the attributes defined, but they are not used once the component is inited. I haven't removed them mostly to make this legacy compatible and in the future we might decide to style components using the attributes.
- Component is controlled by changing it's state, for all _important_ data setters and getters are defined. Updating the state will trigger the attribute update, visual change (re-render) and then event will dispatched (which allows other elements to react to the change).
- New class is added, `PrplGaugeProgressController`, to simplify how gauge and progress bars are synced once the user gets a point or the point is reduced. On the outside still `prplUpdateRaviGauge` function is called in the same way when we want to change points number, so this PR doesn't change other code.

Simplest way to test this, but not 100% accurate regarding side effects, would be to run `prplUpdateRaviGauge( 2 )` or `prplUpdateRaviGauge( -1 )` from the browser console.

But any component can be controlled directly, without affecting other, for example: `document.getElementById( 'prpl-gauge-ravi' ).value = 5`

Progress bars are not longer re-rendered after every change, so the progress slider moves smoothly (using CSS animation). I wanted to the same for the gauge, by using rotation, but in https://github.com/ProgressPlanner/progress-planner-pro/issues/174 we have added gradients for the gauge completed color so that will not work nicely.

Obviously some things can be improved, for example making element attribute names consistent (ie gauge uses value but progress bar points) would make getting through the code much easier and eliminate [this chunk](https://github.com/Emilia-Capital/progress-planner/blob/118a5d5279330019d313c2d0703dfe46bc4030d6/assets/js/web-components/prpl-guage-progress-controller.js#L209-L260).

Also this PR fixes https://github.com/ProgressPlanner/progress-planner/issues/623


aristath (changes requested):
There's something wrong here...

Image

ilicfilip:
Fixed, it was specific to Safari.

aristath (approved):
Fix confirmed. Looks good! 👍

Issue #649, opened 2025-10-03
[Recommendation] PDFs on site recommendations


jdevalk:
### Recommendation title

Improve your site's PDF handling

### Recommendation description

```mermaid
flowchart TD
    A[Start → Does the site have 10 or more PDFs in the media library?] -->|No| Z[No action needed → done]
    A -->|Yes| B[You seem to have quite a few PDFs on your site, let's optimize for that]

    B --> C{Do you need to show a folder structure with the files to make them more discoverable?}
    C -->|Yes| D[Recommend Barn2's Document Library Pro]
    C -->|No| E{Do you want these PDFs<br/>to be found in search engines better?}

    D -->|Install - recommended| E
    D -->|Skip - optional| E

    E -->|Yes| F[Install PDF XML Sitemap plugin]
    E -->|No| G[All good → done]

    F -->|Install - recommended| G
    F -->|Skip - optional| G

    %% Links
    click D href "https://barn2.com/blog/wordpress-pdf-library-plugin/" "Barn2’s Document Library Pro plugin"
    click F href "https://wordpress.org/plugins/pdf-sitemap/" "PDF XML Sitemap plugin on wordpress.org"

```

### Recommendation category

configuration

### Required capability

manage_options

### Onboarding recommendation

false

### Points to award

1

### Recommendation priority

low

### Dismissable

true

### Snoozable

true

ilicfilip:
This was done in https://github.com/ProgressPlanner/progress-planner/pull/650

PR #650, opened 2025-10-03
New task provider: Improve PDF handling

ilicfilip:
Implements https://github.com/ProgressPlanner/progress-planner/issues/649

Things to do:

- Replace placeholder and temporary text
- Improve styling, for example when plugin is installed
- Add data collector for checking the number of PDFs (so query is not ran on every page load)

**EDIT:**
`XML Sitemap for PDFs for Yoast SEO` requires `Yoast SEO` plugin to be active, do we skip suggesting it if `Yoast SEO` is not active or nudge user to activate it as well?

If we do suggest user to activate `Yoast SEO` we probably should check if other SEO plugins are active first, so which ones to include in the check?


aristath:
@ilicfilip what's missing on this one? Is it ready for review, or still a draft?

ilicfilip:
@aristath , decision on:

_XML Sitemap for PDFs for Yoast SEO requires Yoast SEO plugin to be active, do we skip suggesting it if Yoast SEO is not active or nudge user to activate it as well?

If we do suggest user to activate Yoast SEO we probably should check if other SEO plugins are active first, so which ones to include in the check?_

aristath:
@ilicfilip I think both of these were already addressed in https://github.com/ProgressPlanner/progress-planner/pull/676 yesterday 👍

ilicfilip:
I have updated the task so:

- It will be created if user has more than 10 PDF files in the media library
- The first step recommends the premium [PDF Library](https://barn2.com/blog/wordpress-pdf-library-plugin/) plugin
- If the user doesnt have Yoast SEO active there will be no 2nd step and user can collect the point
- If the Yoast SEO is active user can advance to the 2nd step from where it can install / activate the PDF Sitemap add on.

What needs to be done:

- Priority needs to be set correctly, so this task loads after the "Install SEO plugin" task
- Wording needs to be adjusted, as it is mostly Lorem ipsum text now

aristath:
Should this task only be added if the user has Yoast-SEO installed? 🤔 
If they have some other SEO plugin installed this doesn't make a lot of sense

ilicfilip:
Yes, but "The first step recommends the premium [PDF Library](https://barn2.com/blog/wordpress-pdf-library-plugin/) plugin"

So we recommend something in any case

aristath:
@ilicfilip I pushed 2 commits here:
1. Fix inconsistent cache key (`pdf_files_count`, `pdf_count`)
2. Replaced lips text with a reasonable description:
> Adding an XML sitemap for your PDF files helps search engines discover and index them more effectively. This can improve visibility in search results and drive more organic traffic to your valuable PDF content.

PR #664, opened 2025-10-14
Fix secondary button styles

ilicfilip:
Secondary button styles weren't applied properly, the problem was that the background actually comes from the pseudo element and it was coming from the styles for the primary button.


aristath:
Is this one different to #666 ?

ilicfilip:
@aristath , yes - this one corrects styling for the secondary button.

The other PR applies primary button styling to the button in the signup form.

aristath (approved):

PR #667, opened 2025-10-14
Remove categories

aristath:


assets/js/suggested-task.js
```diff
@@ -434,14 +374,8 @@ prplSuggestedTask = {
 			date,
 			date_gmt: date,
 		} );
-		postModelToSave.save().then( ( postData ) => {
+		postModelToSave.save().then( () => {
 			prplSuggestedTask.removeTaskElement( postId );
-
-			// Inject more tasks from the same category.
-			prplSuggestedTask.injectItemsFromCategory( {
```
ilicfilip, line 441: Before we had this called when task was snoozed or completed - another task from the same category was injected to the list.

Now it seems that that is no longer the case, is it on purpose?

classes/update/class-update-190.php
```diff
@@ -24,6 +24,14 @@ class Update_190 {
 	public function run() {
 		// Migrate the golden task.
 		$this->migrate_golden_todo_task();
+
+		// Clean up old category taxonomy data.
+		$this->cleanup_category_taxonomy();
```
ilicfilip: @aristath , what do you think about registering category taxonomy here and then just loop over every term and delete it?

Code is run on `init` hook, which is proper timing for registering taxonomies, also it will be registered only that one time when update class is ran.

assets/js/suggested-task.js
aristath: Good catch! Fixed in https://github.com/ProgressPlanner/progress-planner/pull/667/commits/4a18ea167231002c0adeb2049cec18b4551748d6 👍

classes/update/class-update-190.php
aristath: Nice! Pushed in https://github.com/ProgressPlanner/progress-planner/pull/667/commits/0ea57ef154e87f46cfada2c1601154556d19c58f 👍

classes/update/class-update-190.php
```diff
@@ -42,8 +51,10 @@ private function migrate_golden_todo_task() {
...
 			$points = $wpdb->get_var( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
 				$wpdb->prepare(
-					"SELECT meta_value FROM {$wpdb->postmeta} WHERE post_id = %d AND meta_key = 'prpl_points'", // @phpstan-ignore-line property.nonObject
-					$task->ID
+					'SELECT meta_value FROM %i WHERE post_id = %d AND meta_key = %s',
+					$wpdb->postmeta, // @phpstan-ignore-line property.nonObject
+					$task->ID,
+					'prpl_points'
 				)
 			);
```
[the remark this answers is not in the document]
aristath: False positive

assets/js/widgets/suggested-tasks.js
```diff
@@ -1,10 +1,10 @@
-/* global prplSuggestedTask, prplTerms, prplTodoWidget */
+/* global prplSuggestedTask, prplTerms, prplTodoWidget, prplL10nStrings, history, prplDocumentReady */
```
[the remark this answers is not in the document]
aristath: False positive

PR #668, opened 2025-10-14
Init plugin logic only where needed

ilicfilip:
Most of the code base is instantiated always, both in WP dashboard and on front end although it is not needed.

Long time ago I created made a version of the plugin which doesn't doesn't run code on WP front end if it is not necessary.
Back then I mostly did it for performance reasons and since a lot of things have changed, but I kept the branch in sync with the changes and currently it doesnt have conflicts with the develop branch.

Still in draft state since it needs to be checked in detail.


ilicfilip:
Closing this for now, we can get back to it if needed.

PR #679, opened 2025-10-21
Remove "Recommendations UI" from Playground

ilicfilip:
This PR removes filter which makes the Recommendations UI visible by default from Playground and adds the Toggle button the Debug tools (as it still might be useful to show it sometimes).


aristath (approved):
Looks good 👍

PR #680, opened 2025-10-21
Additional checks and tweaks to strengthen security

aristath:
This PR does the following:

* Prepares queries, removing the hardcoded table-names and uses the `%I` placeholder for them
* Improves sanitisation of values in the settings page
* Adds rate-limiting to REST-API requests to avoid DDOS attacks.

### DIFF
--- a/classes/activities/class-query.php
+++ b/classes/activities/class-query.php
@@ -496,9 +496,11 @@ class Query {
 
 		$table_name = $wpdb->prefix . static::TABLE_NAME; // @phpstan-ignore-line property.nonObject
 
-		foreach ( $wpdb->get_results( "DESCRIBE $table_name" ) as $column ) {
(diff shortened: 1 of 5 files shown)


ilicfilip (commented):
Can we move the `validate_token` method to base class?

IMO transient key can be the same, if authentication fails 10 times it doesnt matter if it failed when `tasks` or `stats` endpoint was pinged?

Also, the test token check (which is the only difference between 2 instances) can be used in both? Originally it was added for tests and `tasks` endpoint is only used in that case (if I am not mistaken).

If I nitpick, I dont think the change in `class-query.php` was needed as $table_name was practically hard coded, but could be that I am missing something.

aristath:
Fair points! Pushed tweaks.

As for the class-query.php change, it doesn't hurt... and preparing is best practice. I agree that it won't have an impact so it's just a "useless" change, but if you don't see any negative impact I'd prefer to keep it 👍

PR #663, opened 2025-10-13
Convert more tasks to Interactive

ilicfilip:
- fewer-tags
- core-permalink-structure
- rename-uncategorized-category
- search-engine-visibility
- core-siteicon
- remove terms without posts
- update term description
- Yoast: author archive
- Yoast: remove emoji scripts
- Yoast: remove author feeds
- Yoast: remove global comments feeed
- Yoast: media pages
- Yoast: organization logo (besides logo user / organization data has to be set)

PR #683, opened 2025-10-24
v1.9 dev - Fix interactive tasks

ilicfilip:
Interactive tasks were broken in multiple ways:

- In https://github.com/ProgressPlanner/progress-planner/pull/586 we added a way to handle failed request for interactive tasks, but in https://github.com/ProgressPlanner/progress-planner/pull/663 we converted new tasks to be interactive. The tasks which were added later needed to be adjusted.
- `wpseo` was missing from the `allowed options to update` list
- `remove terms without posts` needed to be adjusted for the new way of how we store `task_id`

I started fixing one issue, which @tacoverdo reported, but then while testing I discovered more and fixed them - but another round of interactive tasks testing would be good.


aristath (approved):
Code looks good and works fine.

PR #628, opened 2025-09-18
v1.9.0

aristath:

PR #685, opened 2025-10-24
Add spinner (loader) to the interactive tasks forms

ilicfilip:
Inspired by [this comment](https://github.com/ProgressPlanner/progress-planner/pull/628#issuecomment-3442109498)

Interactive tasks look similar, almost the same, but their HTML structure and the way they are submitted is different.
One example of different HTML structure is how submit button is positioned, for tasks which have multiple steps it is in the bottom right corner. But tasks which only have that button (like "Disable comments") will have it in the top left.

Spinner needs to be added to the side of the button which has available space (so the button is not moved when spinner is visible) and I have tried not to change how buttons were positioned before, but because there are many different variations another round of testing would be good.

### DIFF
--- a/assets/css/page-widgets/suggested-tasks.css
+++ b/assets/css/page-widgets/suggested-tasks.css
@@ -441,9 +441,21 @@
 			padding-top: 1rem;
 			display: flex;
 			justify-content: flex-end;
+			align-items: center;
 			gap: 1rem;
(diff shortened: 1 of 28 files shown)


tacoverdo (approved):
Tested in Playground and on a live website. The spinner is working on either the left or right side of the button, as intended.

PR #686, opened 2025-10-24
Add spinner to onboarding form and remove success message

ilicfilip:
This PR removes the success message from the onboarding form. In the past there, in the background, site content was scanned but it doesnt serve much purpose since that was removed.

Besides that spinner is added to the onboard form once it is submitted, also adding the disabled attribute is corrected (so now it is added to both submit buttons) so user knows something is happening.

### DIFF
--- a/assets/css/onboard.css
+++ b/assets/css/onboard.css
@@ -16,3 +16,8 @@
 		display: block !important;
 	}
 }
+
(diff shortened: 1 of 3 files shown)


tacoverdo:
Yes! This is infinitely better in user experience!
Thanks, @ilicfilip!

PR #689, opened 2025-10-27
Fix "check irrelevant tasks" for "Delete term without posts" and "Update term description" tasks

ilicfilip:
While testing for v1.9 this morning @tacoverdo and I noticed that a task for "Delete term without" posts, which was created properly, gets marked as completed without term actually being deleted.

After some debugging I found that it actually happens because `maybe_remove_irrelevant_tasks` methods were called for all PP specific (not public taxonomy).

Besides that not being needed, since we're interested only in the public taxonomies, it also created a bug because the taxonomies are assigned before the post meta - and we need post meta in order for `$term` to be fetched [here](https://github.com/Emilia-Capital/progress-planner/blob/959b6e0053f7da65ced37e53cde3a2db4b2eed40/classes/suggested-tasks/providers/class-update-term-description.php#L186).

Since post meta was missing, `$term` had a value of `NULL` and the recommendation was marked as completed.

This PR fixes that issue.


tacoverdo:
It no longer shows the 'weird' behavior around taxonomies, so this patch fixes the problem.

PR #696, opened 2025-10-30
Reduce code duplication

aristath:


classes/rest/class-base.php
```diff
@@ -12,6 +12,13 @@
  */
 abstract class Base {
 
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		\add_action( 'rest_api_init', [ $this, 'register_rest_endpoint' ] );
```
ilicfilip, line 19: What do you think about adding`abstract public function register_rest_endpoint()` as well?

This way child class will trigger the error if it doesnt implement it or override the constructor.

aristath: Good catch! Added in https://github.com/ProgressPlanner/progress-planner/pull/696/commits/e31cbd220a324a7d49eb96a339da582414b744f7 👍

ilicfilip (approved):

PR #697, opened 2025-10-30
Reduce code duplication using traits

aristath:


aristath:
Note for PHPUnit coverage: It's impossible to get proper coverage on code that calls `wp_send_json_*()` and exits. This is a fundamental and known limitation of PHPUnit Coverage reports.
However, the simplifications in this PR and the additional PHPUnit tests result in a 1.7% increase in total coverage, so overall the PR is in excellent shape 🎉

ilicfilip (approved):
I like it!

PR #698, opened 2025-10-30
Code modernization

aristath:
This PR applies some code modernization since our minimum required PHP version is 7.4, and we don't support OLD browsers.

* Add typed props and param types
* `progressPlannerAjaxRequest` now uses the Fetch API
* Use array speed operator instead of `array_merge`
* Other minor code modernizations


aristath:
Closing this one, it causes too many issues. We can revisit at a later date.

PR #699, opened 2025-10-31
Improve inline documentation

aristath:
This was a big one...

Doesn't change any actual code, but we increase code-quality by documenting lots of things that were previously ambiguous.

### DIFF
--- a/classes/class-badges.php
+++ b/classes/class-badges.php
@@ -170,7 +170,6 @@ class Badges {
 	 * - Later completion dates take precedence (>= comparison ensures newer badges win)
 	 *
 	 * This is used to:
-	 * - Display the most recent achievement on the dashboard
 	 * - Trigger celebrations for newly completed badges
 	 * - Track user progress momentum
 	 *
--- a/classes/class-badges.php
+++ b/classes/class-badges.php
@@ -156,35 +156,52 @@ class Badges {
 	}
 
 	/**
-	 * Get the latest completed badge.
+	 * Get the latest completed badge across all badge types.
 	 *
-	 * @return \Progress_Planner\Badges\Badge|null
+	 * Badge selection algorithm:
+	 * 1. Iterates through all badge contexts (content, maintenance, monthly_flat)
+	 * 2. For each badge, checks if it's 100% complete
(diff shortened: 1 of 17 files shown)


classes/suggested-tasks/providers/class-tasks.php
```diff
@@ -267,20 +267,37 @@ public function get_external_link_url() {
...
-	 * @param array $task_data Optional data to include in the task ID.
-	 * @return string
+	 * Generates a unique task ID by combining the provider ID with optional task-specific data.
+	 * For repetitive tasks, includes the current year-week (YW format) to create weekly instances.
+	 *
+	 * Example task IDs:
+	 * - Non-repetitive: "update-core"
+	 * - With post target: "update-post-123"
+	 * - With term target: "update-term-5-category"
+	 * - Repetitive weekly: "create-post-2025W42"
```
ilicfilip, line 277: I think this should just be "create-post-202542"

classes/class-badges.php
```diff
@@ -156,35 +156,52 @@ public function clear_content_progress() {
...
+	 * 3. Compares completion dates stored in settings to find the most recent
+	 * 4. Returns the badge with the most recent completion date
+	 *
+	 * The completion date is stored in settings when a badge reaches 100% progress:
+	 * - Format: 'Y-m-d H:i:s' (e.g., '2025-10-31 14:30:00')
+	 * - Compared as Unix timestamps for accurate chronological ordering
+	 * - Later completion dates take precedence (>= comparison ensures newer badges win)
+	 *
+	 * This is used to:
+	 * - Display the most recent achievement on the dashboard
```
ilicfilip: We removed the latest badge from the Dashboard?

ilicfilip (approved):

classes/suggested-tasks/providers/class-tasks.php
aristath: Yep, you're right. Fixed 👍

classes/class-badges.php
aristath: Good catch!

PR #425, opened 2025-04-14
Check remote API status

ilicfilip:
## Context

We ping our API for a lot of things, not just for registration and getting data but also getting badge graphics and etc.

We have fallback for images and we cache API response for 5 minutes if the response from our server was empty, but users don't know that which can be confusing.

This PR adds following:

- A note on the Settings page if the Page types are not loaded from API
- A button on the Settings page which pings our server and displays a status message (so user can check API status in the real time).
- Improves check for "Save settings" task, to add it only if Page types are fetched from our server

The thing that still needs to be implemented is some kind of notice, probably on the PP Dashboard page when it is detected that some API call has failed.

Fixes https://github.com/ProgressPlanner/progress-planner/issues/422

PR #702, opened 2025-11-03
Convert Settings page to interactive tasks

ilicfilip:
This PR converts settings from Settings page to interactive tasks, as much as possible.

It is still WIP but set About, Contact and FAQ page settings are converted - the choices user has are the same:

- I have this page
- I dont have this page yet (Created this page)
- My site doesnt need this page

All 3 tasks share same codebase, but we can adjust popover to recommend user to install Contact form or any other feature.

Things left to implement (from Slack discussion):

- ~Detect if site has home page set and auto assign it (no need for user to do it)~ We already [do this](https://github.com/Emilia-Capital/progress-planner/blob/423a1fc03d607c2167d1bc79322d0097a0136432/classes/class-page-types.php#L361-L393)
- Convert "Select the Valuable Post Types" to interactive task, it should reappear every time a new post type is added and/or removed


ilicfilip:
The "Set valuable post types" task is also converted to be interactive.

Task will be added again if the new public post type is detected. I think there is no need to add it when public post type is removed since that means there will just be one less checkbox to be selected, which doesn't give user any other option than confirming the current selection (also it might be confusing, if user doesn't realize that one post type is now missing).

Since "Set pages" and "Set valuable post types" are now interactive tasks the only option on that page is "Default login destination". If we move it to the PP dashboard page, maybe create a button (next to the Tour and Registration) which triggers popover with that option?

If we do that, should we remove Settings Page completely and "Fill the Settings Page" task as well?

EDIT:
There was an idea about implementing a way for user to check if our API is available ( https://github.com/ProgressPlanner/progress-planner/pull/425 ), but it was never finished. I think it's a good thing to have, since we are increasingly dependent on our server, but we will need to find a new place for this if the Settings page is removed.

aristath (approved):

aristath:
@ilicfilip In this PR we need to:
1. Remove the admin page
2. Remove the task to fill-in the settings page.

Once these get added, this will be ready to merge 👍 🥇

PR #705, opened 2025-11-06
Integrate with Elementor's Angie

ilicfilip:
This PR implements https://github.com/ProgressPlanner/progress-planner-pro/issues/182 (still WIP of course)

Technically implementation is described in [README.md](https://github.com/Emilia-Capital/progress-planner/blob/4ef4ec77f9150ed394402b1405e9a3143f089916/classes/third-party/angie/README.md#L1), so I will focus more on testing initial implementation and what needs to be done in the future.

When Progress Planner MCP server is registered Angie becomes aware of it and it is smart when you ask her about it, for example queries like "List me all (completed) Progress Planner tasks" work without the a problem.

Tasks like "Can you complete 'Set site tagline' Progress Planner task for me? I would like to set tagline to 'Thank you!'" work fine most of the time. 

Good example is "Can you complete the "Set site timezone" Progress Planner task? I would like to set the timezone to Amsterdam" as well. It is smart enough to pass Europe/Amsterdam to our API callback so the `timezone_string` `wp_option` is correctly set.

Completing the "Set site locale" didnt work for me, what Angie did was that it set the site locale (by setting the `wp_option` directly) and then tried to mark PP task as completed through our API endpoint. That is the thing which we need to be careful when testing, as PP plugin codebase will mark task as completed if the condition is satisfied, but it doesnt mean that it was completed through the Angie integration itself. More on that below.

Also my tries to tell it to complete the "Set the site icon" task to the icon which I have uploaded failed, it didn't even add icon to the Media library

Angie runs in an iframe and the WP Dashboard on the right doesn't update without refreshing it. Mentioning it since it will look a bit weird if Angie says that she changed site locale but change is not visible right away. This can be especially be weird if we trigger confetti (yes, it should be possible to do some kind of celebration at least) but in order to do so we need to make sure that tasks are completed through our integration (not by setting `wp_options` directly).

That can probably be improved by providing better context to Angie [here](https://github.com/Emilia-Capital/progress-planner/blob/4ef4ec77f9150ed394402b1405e9a3143f089916/classes/third-party/angie/src/progress-planner-mcp-server.ts#L152-L156).

When testing be aware that you have (only) 900 tokens per day on a free plan.

One technical note, code is placed in this way so our autoloader works (at first I placed it in `./third-party/angie`, but autoloader requires classes only from classes directory) so we will probably move it to a better place.


ilicfilip:
Notes from the recent commits:

- Tasks which Angie 'sees' are limited now, she can see only the tasks which she can complete: core-blogdescription, set-locale, select-timezone (WIP of course, the list will be expanded).
- Angie tool definition is moved to PHP, so JS doesnt need to be rebuilt every time there is a change (new tool is added or simply description changed). This will be further improved.
- JSON schema, which is used to define Angie tool, is improved (uses more strict type checking) and expanded.

We should be careful when defining schema, docs are not that clear and a small mistake can cause Progress Planner MCP server (or tools) not to be registered.

ilicfilip:
## Tested against a live Angie 1.1.12 install

Notes from bringing this branch up to date, since a couple of things here are easy to miss.

### It works now, but it needed a registration change

Angie 1.1.12 discovers MCP-capable plugins through the `angie_mcp_plugins` filter and passes the result to its app as `window.angieConfig.plugins`. This branch never added itself to that list, so the MCP server script it enqueues registered into nothing — confirmed on a live install, where the filter returned only `["elementor_pro"]`.

Added in 21ac0b80, matching how Angie's own Elementor module registers. After that, Angie discovers the tools and calls them: the composer shows a **Progress Planner** badge and the task list renders with action buttons.

### The "you have no active tasks" bug

`get_tasks_by_status()` returned **only** the three recommendations Angie can complete itself (`core-blogdescription`, `select-locale`, `select-timezone`) and silently dropped everything else.

On a live site that meant:

```
published tasks in DB:  16
endpoint returned:       0
```

None of those 16 happened to be one of the three, so Angie reported *"You currently have no active tasks in Progress Planner"* on a site with 16 open recommendations. Even once two of the three existed, it listed 2 of 19 with no indication the rest were hidden.

Fixed in 8789e429: every recommendation is listed, each carrying a `can_complete` flag. This does **not** widen what Angie can change — `complete_task()` enforces the same map — it just stops the list from lying about what exists.

### A second bug found while fixing that one

`complete_task()` fell through to `$task->celebrate()` for *any* task ID not in the map. So asking Angie to complete something it has no handler for (e.g. `hello-world`) would mark the recommendation done **without touching the site it describes** — the user is told their site changed when nothing did.

It now returns `task_not_completable` and leaves the recommendation open.

### The provider API had moved out from under this branch

Worth flagging for anyone picking this up: develop's onboarding rework (39a2e0da7) removed `complete_task()` from the task providers. `Angie_API::complete_task()` still called `$task_provider->complete_task()`, which is a **fatal on every Angie task completion**.

The replacement (`handle_interactive_task_submit()`) reads `$_POST` and terminates with `wp_send_json_*()`, so it cannot be called from a REST callback that has to return a response. The update now happens in `Angie_API::update_task_setting()`, which reuses the provider's option whitelist rather than bypassing it — the old `complete_task()` path had no such restriction, so this is slightly tighter than before.

Also: three test files (`select-locale`, `select-timezone`, `core-blogdescription`) existed in both branches. This branch's versions were larger and arguably better designed, but all targeted the removed API and errored, so develop's versions were kept.

### Still open

- **Path B — the Abilities API.** Angie 1.1.12 also exposes tools via `wp_register_ability()` (its `wp-abilities` module), which would be consumer-agnostic and could replace the ~500-line TS source and its ~12k built bundle entirely. Not viable to develop locally as things stand: Angie's `.org` build ships without `vendor/`, so `WP\MCP\Core\McpAdapter` never loads and the module stays inactive.
- **Completion flow is unverified end-to-end through Angie.** The REST layer is tested (whitelist enforced, timezone mapping correct, real options written), but nobody has driven an actual task completion through the chat UI since the refactor. The original description notes "Set site locale" behaving oddly — worth re-checking.
- The two root-level docs (`AI_TASKS_*.md` style working notes) and the committed `dist/` bundle are still there.

ilicfilip:
## Notes: porting to the Abilities API (Path B)

Recording the design discussion so it isn't lost. Nothing here is implemented — this is what a port would involve if we continue the Angie integration.

### Why bother, when the current approach works

The integration works today (see previous comment): Angie discovers the MCP server via `angie_mcp_plugins` and calls its tools.

The argument for moving is that completion should go **through plugin code**, not around it. For a side-effect-free task like the tagline, `update_option( 'blogdescription', … )` and "completing the task" are literally the same write — so if Angie ever set it directly, the outcome would look identical while proving nothing about the integration. That's a false green, and it matters for the complex tasks where the two *aren't* equivalent.

Framing the simple tasks as a canary for the integration is the right instinct. It only works if the completion demonstrably routes through our code.

### What already exists on this branch

Worth stating, because analysis done against `develop` will conclude otherwise:

- **A REST completion route exists** — `POST /progress-planner/v1/angie/recommendations`. It's the REST sibling of the admin-AJAX `handle_interactive_task_submit()`, running the same option allowlist and capability check. `develop` has no such route, so a survey of `develop` will report this as missing.
- **Tool discovery works** — via the `angie_mcp_plugins` filter (21ac0b80).

So the "we need a REST equivalent of the AJAX handler" step is already done. What's missing is only the ability registration layer.

### What a port would add

Following Angie's own pattern (`Mcp_Adapter_Ability_Registration::register_rest_tool()` wraps an existing REST route as an ability):

- `progress-planner/list-tasks` → wraps the existing `GET /angie/recommendations`
- `progress-planner/complete-task` → wraps the existing `POST /angie/recommendations`

Both routes already exist and are tested, so this is a registration layer over them rather than new endpoint work.

The payoff is that abilities are **consumer-agnostic**: they'd serve Angie and any other MCP client, and would let us delete the ~500-line TS source and its ~12k committed `dist/` bundle. Angie's app is fetched from a remote bucket, so the current SDK-bundle approach is coupled to a client contract we can't see or pin.

### Where to register it

Progress Planner registering its own, guarded by `function_exists( 'wp_register_ability' )` — no hard dependency for older WP, and the capability travels with the feature rather than depending on Angie's module layout.

### Blocker for local work

Angie's `.org` build ships **without `vendor/`**, so `WP\MCP\Core\McpAdapter` and `WP\MCP\Transport\HttpTransport` never load and `Wp_Abilities_Support::is_mcp_adapter_supported()` returns false — the whole `wp-abilities` module stays inactive.

Checked on this install:

| | |
|---|---|
| WP 7.1 | ✅ |
| `wp_register_ability()` | ✅ |
| `WP\MCP\Core\McpAdapter` | ❌ |
| `WP\MCP\Transport\HttpTransport` | ❌ |

So this can't be developed or tested locally as-is. Needs either an Angie build with its Composer deps, or `wordpress/mcp-adapter` pulled in separately.

### Separate bug worth fixing regardless

`Tasks_Interactive::handle_interactive_task_submit()` on `develop` treats an unchanged value as failure:

```php
$updated = \update_option( $setting, $value );
if ( ! $updated ) {
    \wp_send_json_error( [ 'message' => … 'Failed to update setting.' ] );
}
```

`update_option()` returns `false` when the new value equals the old one, so re-submitting a setting that already matches reports a failure that isn't one. That affects the human popover path today, independently of Angie.

The REST route on this branch doesn't inherit it (it doesn't check the return value), but the two should agree on what "success" means — worth fixing on `develop` rather than only here.

PR #718, opened 2025-12-02
Recognize SureRank as an SEO plugin

ilicfilip:
Implements: https://github.com/ProgressPlanner/progress-planner-pro/issues/186

cc @tacoverdo


jdevalk:
This LGTM @ilicfilip - don't get the CS fixes but I'm assuming you can do those :)

PR #725, opened 2025-12-15
Fix WP CLI get task command & add tests

ilicfilip:
WP CLI get task command wasnt working after recent refactors, this PR fixes it and adds tests for following WP CLI commands (if WP CLI is available):

```
  | test_task_list                           | Lists tasks in JSON format 
  | test_task_list_table_format              | Lists tasks in table format             |
  | test_task_create                         | Creates a task with required fields     |
  | test_task_create_without_required_fields | Verifies error without task_id/title    |
  | test_task_get                            | Gets a single task by ID                |
  | test_task_get_not_found                  | Verifies error for non-existent task    |
  | test_task_update                         | Updates a task's status                 |
  | test_task_delete                         | Deletes a task and verifies removal     |
  | test_full_crud_cycle                     | Complete create/read/update/delete flow |
```


ilicfilip:
I have changed approach:

Initially WP CLI commands were triggered like they are triggered from the command line, using exec. This worked fine locally (since it actually ran in local repo, not the test WP instance) but didnt in CI. To make it work I had to symlink repo to the test WP instance, create wp-config.php, install WP core, activate the plugin, change wp table prefix..

Now the WP CLI commands are tested from PHP directly, not using the exec command. This doesn't require any of the above steps and it is much faster. Also it feels more natural since.. this are PHP Unit tests.
The only setup change is that WP CLI is now required as dev composer dependency.

aristath (approved):

PR #727, opened 2025-12-15
Delete user meta, which redirects to PP dashboard page

ilicfilip:
In https://github.com/ProgressPlanner/progress-planner/pull/702 we have removed the Settings page, with it the "Redirect to PP Dashboard on login" setting is gone as well - and users cant prevent from being redirected.

This PR adds an update class which deletes user meta which was used as a flag for redirection.

### DIFF
--- a/classes/update/class-update-1100.php
+++ b/classes/update/class-update-1100.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Update class for version 1.10.0.
+ *
+ * @package Progress_Planner
+ */
+
+namespace Progress_Planner\Update;
+
+/**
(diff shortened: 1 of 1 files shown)


ilicfilip:
Follow-up for the 1.10.0 release: the redirect-on-login handler, its `wp_login` hook, and the unused `save_redirect_on_login()` setter were still in the code after the Settings page was removed here. They were no longer reachable (the meta this migration deletes was the only thing that enabled the redirect, and nothing writes it anymore), so they've been removed as dead code in #780. This migration stays — it still cleans up the historical user meta.

PR #728, opened 2025-12-16
When we're acting as a host, move the menu item to position 0

jdevalk:
So we're the top menu item.


ilicfilip (approved):
It works

@jdevalk , what do you think about adding a filter for the position, so it can be used in pp-hosts plugin?

that way if we need to tweak something we can just release new pp-hosts version, instead of making a new PP release

aristath (approved):

PR #730, opened 2025-12-17
New onboarding tweaks

ilicfilip:
This PR:

- Removes "Default login destination" checkbox from onboarding, since we removed that option in https://github.com/ProgressPlanner/progress-planner/pull/702
- Adds `max-width` to onboarding logo, to prevent horizontal overflow
- Changes condition for when Upgrading tasks should be displayed, since we changed when privacy policy was accepted for branded instances in https://github.com/ProgressPlanner/progress-planner/pull/726

### DIFF
--- a/assets/css/onboarding/onboarding.css
+++ b/assets/css/onboarding/onboarding.css
@@ -298,6 +298,7 @@
 		img, svg {
 			height: 100%;
 			width: auto;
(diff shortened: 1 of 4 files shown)


ilicfilip:
Merging this one as it will be a part of new pp-hosts release.

PR #734, opened 2025-12-24
Leverage WordPress 6.9+ Notes for Review Tasks

ilicfilip:
This PR integrates with WordPress 6.9's new Notes feature to make "Review post X" tasks more actionable. Instead of just sending users to the edit page, we now create block-level notes that guide them through reviewing specific elements.
  
Implements: https://github.com/ProgressPlanner/progress-planner-pro/issues/192

**Changes**

  - Added "Review with Notes" action button on review tasks (WP 6.9+ only)
  - When clicked, automatically injects notes for:
    - Images: Each core/image block gets a note asking to verify relevance and display
    - Linked images: Special message noting both image and link should be checked
    - Paragraphs with links: One note per paragraph containing links, with link count
  - Notes are linked to blocks via metadata.noteId attribute
  - Task completion now checks if all PRPL notes are resolved (when notes exist)
  - Falls back to original "post modified" check when no notes present
  - Duplicate prevention: won't create notes if block already has a valid note

**How It Works**

  1. User clicks "Review with Notes" on a review task
  2. Opens editor with ?prpl_inject_notes=1 parameter
  3. Notes are created and linked to relevant blocks
  4. User sees notes in the Notes panel, can click to highlight blocks
  5. User resolves notes as they review each element
  6. When all notes resolved → task auto-completes

**Testing**

  1. Create a post with images and paragraphs containing links
  2. Backdate post_modified to >12 months ago
  3. View the review task and click "Review with Notes"
  4. Verify notes appear in the Notes panel
  5. Resolve all notes and verify task completion


tacoverdo:
Pushed 3a5bbe3d, which makes the injected notes visible right away: clicking "Review with Notes" now auto-opens the Notes sidebar in the editor.

**What it does:**
- New `assets/js/review-post-notes.js` waits for the block editor to finish initializing (so the sidebar restored from user preferences doesn't win), then opens the "All notes" sidebar (`edit-post/collab-history-sidebar`) via `enableComplementaryArea`.
- `Content_Review::maybe_enqueue_notes_sidebar_script()` (hooked to `enqueue_block_editor_assets`, inside the existing `supports_notes()` gate) enqueues it only when `prpl_inject_notes=1` is present, the user can edit the post, and the post has at least one **open** `[PRPL]` note — so posts with nothing to flag, or with everything resolved, don't get an empty sidebar opened. Since `load-post.php` runs before asset enqueueing, freshly injected notes are already there when this check runs.
- Added a PHPUnit test covering all three branches (no param / no open notes / open notes).

**Why the explicit dispatch:** core only auto-opens the floating notes sidebar when no other sidebar is active — most users have the Document panel open, which suppresses it. Targeting the pinned "All notes" sidebar also works in all rendering modes, unlike the floating one (post-only + large viewports).

**Tested:** full PHPUnit suite green against WP 7.1, 7.0.4, and 6.9.7 (416 tests); on 6.7.7 the notes tests correctly stay inert (1 expected skip). WP 6.5 predates the plugin's 6.6 minimum and fails at bootstrap for unrelated reasons (`wp_autoload_values_to_autoload()`).

PR #743, opened 2026-01-13
Add null guards to prevent fatal errors during pp-hosts plugin updates

ilicfilip:
When pp-hosts (or other host plugins bundling Progress Planner) is updated, hooks registered by the old plugin version can fire during the update process. At that point, dependencies may not be available because files are being extracted, causing fatal errors like:
  - `Call to a member function get_post_types_names() on null`
  - `Class "Progress_Planner\Activities\Maintenance" not found`

This PR fixes that by adding checks and making early exits:

- Add null check for `get_activities__content_helpers()` in Content class `should_skip_saving()` method
- Add null checks in Content_Scan class `get_total_pages()` and `update_stats()` methods
- Add `class_exists()` check in Maintenance class before instantiating `Activities_Maintenance`

### DIFF
--- a/classes/actions/class-content-scan.php
+++ b/classes/actions/class-content-scan.php
@@ -59,12 +59,18 @@ class Content_Scan extends Content {
 			return;
 		}
 
+		// Bail if content helpers are not available (can happen during plugin updates).
+		$content_helpers = \progress_planner()->get_activities__content_helpers();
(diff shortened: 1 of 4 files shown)


ilicfilip:
Updating through pp-hosts was tested, there are no errors any more so I am merging this one.

